PR #10, opened 2025-02-14
Added Logo image and logo blender file.

michael-paulin:
Added the Logo image and logo blender file.


seehorne (changes requested):
Do we think we could make a subdirectory for these, like maybe image_utils and put them in there? i anticipate we'll have a lot more visual elements over the course of the semester and it would be good to start getting them organized now

almondheil:
We'll also need to put an explanation of that dir in the layout part of the README.

seehorne:
Yeah that's fine, I can make the relevant changes in the readme @almondheil

almondheil (approved):
Should be good with moved logo files

seehorne (approved):

PR #12, opened 2025-02-14
Update README.md

seehorne:
update readme to reflect existence of image-utils


almondheil (approved):
Looks good! In future, we might want to also add this to the same PR to avoid having multiple pulls over it.

AnthonyS25223 (approved):
Updates look good to me

PR #20, opened 2025-03-04
Ignore .idea dir, it's only metadata

almondheil:
This PR brings main up to date with the Kotlin testing branch, including deleting the .idea directory from the repo. After this request is merged, we SHOULD NOT DELETE THE SOURCE BRANCH.


seehorne (approved):

AnthonyS25223 (approved):
Looks good

PR #22, opened 2025-03-17
Create docs dir, set up automatic build/deploy through workflow

almondheil:
This pull request sets up MkDocs for hand-written documentation, which will be where we can host some of our user guides and materials as detailed in the documentation plan. It also creates a Github action that will run when the doc files are changed by a push, which deploys them to our site.

# Note to reviewers

All actual documentation content is in `mkdocs.yaml` and `docs/`. The actual workflow is in `.github/workflows/mkdocs.yml`.

Currently, the docs are only a note saying that there's nothing there.

Ellie and I have worked together to test the workflow and make sure it's all okay.

### DIFF
--- a/.github/workflows/mkdocs.yml
+++ b/.github/workflows/mkdocs.yml
@@ -1,0 +1,28 @@
+name: Build MkDocs
+
+on:
+  # Run on pushes to specified branches.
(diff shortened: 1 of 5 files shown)


AnthonyS25223 (approved):
The files look good, thank you for your amazing work!

seehorne (approved):

PR #33, opened 2025-04-01
Run when any backend code is pushed

seehorne:
I'm pretty sure this will fix it, can someone who knows more check that I'm right and then approve?


almondheil (commented):
We should have this set up now. Thank you! I'll close this request since we should be all set.

See https://github.com/seehorne/GetGrinnected/commit/fcae048d6d3427dd31237c9dfde8f72b4fd7c414

PR #35, opened 2025-04-02
Pull sprint 3 planning into main.

michael-paulin:


AnthonyS25223 (approved):
Report looks good (since it was made collaboratively by all members in class I think we can just approve)

written reports/sprint 3/agile-planning.md
```diff
@@ -0,0 +1,125 @@
+# Sprint 3 planning
+
+## Capacity
+
+### Michael: 3.5 blue beans
```
almondheil, line 5: I love the different types of beans! This is a good way of keeping this portion light-hearted, we should continue to do so in the future.

almondheil (approved):
There's one comment about something we could improve for the future, but since this was already due we probably shouldn't make changes--I don't believe sprint plannings are intended to be living documents.

PR #36, opened 2025-04-05
Final Development Merge

AnthonyS25223:
These are the final edits to the development branch before we chose to switch to a different branching style.
- It Includes work on the kotlin UI and navigation.


src/android/GetGrinnected/MyApplication/app/src/androidTest/java/com/example/MainActivityKtTest.kt
```diff
@@ -16,23 +16,23 @@ class MainActivityKtTest {
 
  @Test
  fun welcomeScreen_LoginButton_NavigatesCorrectly() {
-  composeTestRule.setContent { AppNavigator() }
+  composeTestRule.setContent { AppNavigation() }
```
almondheil, line 19: What does this change do? Is it about the back gesture behavior we talked about?

src/android/GetGrinnected/MyApplication/app/src/main/java/com/example/myapplication/AppNavigation.kt
```diff
@@ -0,0 +1,38 @@
...
+import screens.SignupScreen
+import screens.WelcomeScreen
+
+/**
+ * A composable function that is utilized for smooth navigation through login/signup process
+ * @param modifier the modifier to be applied to page layouts navigated to.
+ */
+@RequiresApi(Build.VERSION_CODES.O)
+@Composable
+fun AppNavigation(modifier: Modifier = Modifier){
```
almondheil, line 20: OH! We are redefining our own navigation to call composables for each screen, rather than just updating a "current screen" variable. Very good.

src/android/GetGrinnected/MyApplication/app/src/main/java/com/example/myapplication/MainPage.kt
```diff
@@ -33,53 +40,51 @@ import androidx.compose.ui.Modifier
...
- * @param selectedIndex The index of the currently selected navItem.
- */
-@Composable
-fun ContentScreen(modifier: Modifier = Modifier, selectedIndex : Int){
-    when(selectedIndex){
-        0 -> HomeScreen()
-        1 -> CalendarScreen()
-        2 -> FavoritesScreen()
-        3 -> SettingsScreen()
+    ) { innerPadding ->
```
almondheil, line 78: This is a much more robust way to implement it. That addition of `innerPadding` also seems to fix the issues I had with overlaps before, so great job!

almondheil (approved):
Everything looks good to me! I've been able to test them on my phone and things are looking great.

This wasn't changed in this PR (and I'll talk to Ethan about it), but with the fixes to the margins there is now a visible difference between the pure-black bar at the top of the home screen and the dark-gray default background.

src/android/GetGrinnected/MyApplication/app/src/androidTest/java/com/example/MainActivityKtTest.kt
```diff
@@ -16,23 +16,23 @@ class MainActivityKtTest {
...
 
   composeTestRule.onNodeWithText("Login").performClick()
-  composeTestRule.onNodeWithText("Login to you account").assertIsDisplayed()
+  composeTestRule.onNodeWithText("Login to your account").assertIsDisplayed()
```
seehorne, line 22: this test appears to be failing even though the typo is fixed

src/android/GetGrinnected/MyApplication/app/src/androidTest/java/com/example/MainActivityKtTest.kt
```diff
@@ -16,23 +16,23 @@ class MainActivityKtTest {
...
  }
 
  @Test
  fun welcomeScreen_SignUpButton_NavigatesCorrectly() {
```
seehorne, line 26: this is a helpful test name, good job being explanatory

src/android/GetGrinnected/MyApplication/app/src/androidTest/java/com/example/MainActivityKtTest.kt
AnthonyS25223: The naming convention was more on me, but yes the app navigation is now a navigator composable that we can use to have more standard back gestures to what you would expect.

src/android/GetGrinnected/MyApplication/app/src/androidTest/java/com/example/MainActivityKtTest.kt
```diff
@@ -6,7 +6,7 @@ import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
 import androidx.test.ext.junit.runners.AndroidJUnit4
```
seehorne, line 8: it seems like maybe this won't run directly/standalone, that might be the CI problem

seehorne (changes requested):
Good work, but CI doesn't run. Almond and I speculate that it might be that this file doesn't run standalone and Github actions/gradle don't really know how to get that to happen without information that would be available locally but not on Github server. Seems to run locally though, so it will ultimately be a convince Github problem

AnthonyS25223:
We decided as a group to disable the test for the time being to allow additional progress in Kotlin using the new branching framework we hope to follow. Additionally, we made an issue to address this failing test in the CI at a later point.

seehorne (approved):
Ok, I see what we've decided to do with the test, understandable, feel free to push

PR #37, opened 2025-04-06
Add getEventsBetween functionality to API, test API

almondheil:
There's a lot in this pull request because there was a lot of work to do in order to get the API to a workable state. Here is a guide to reviewers by file.

* `.github/workflows/node.yaml`
  * Run on all branches, since we're getting rid of `development`
  * Run `npm test` instead of `npm run tests` since it's more common.
    * Still runs all the same tests, don't worry
* `src/backend/api.js`
  * make `run()` and `close()` functions that let you start/stop the API.
  * reorganize routes into their own functions
  * add /events/between/START/END route
  * write helpers to parse date params and tag query params
* `src/backend/db_connect.js`
  * create functions to get events between dates (tagged and not tagged)
  * make helpers that reduce repetition for making a query over all tags and a query between two dates
  * TODO: all-day events still need to be handled
* `src/backend/test/test_api.js`
  * test everything I implemented
  * useful tools: supertest for API testing, sinon for faking methods (to avoid hitting DB)

### DIFF
--- a/src/backend/db_connect.js
+++ b/src/backend/db_connect.js
@@ -83,7 +83,7 @@ async function getEvents(){
  * \returns     A string that serves as a SQL selector for all those tags
  */
 function queryAllTags(tags) {
-    const pieces = tags.map((t) => `JSON_CONTAINS(tags, '"` + t + `"', '$')`);
+    const pieces = tags.map((t) => `JSON_CONTAINS(tags, '"` + pool.escape(t) + `"', '$')`);
     const piecesAndTogether = pieces.join(' AND ');
     return '(' + piecesAndTogether + ')';
 }
@@ -98,7 +98,7 @@ function queryAllTags(tags) {
 function queryBetweenDates(start, end) {
     // events where the start and end times are set
     // TODO: comparisons seem to not be working. need more work
-    const event_time_between = `(event_start_time < FROM_UNIXTIME(${start}) AND event_end_time > FROM_UNIXTIME(${end}))`
+    const event_time_between = `(event_start_time < FROM_UNIXTIME(${pool.escape(start)}) AND event_end_time > FROM_UNIXTIME(${pool.escape(end)}))`
 
     // TODO: handle events that are marked all day and therefore don't have a start time and end time
     // could it be done when scraping, maybe? this will SUCK to do in mysql.
--- a/src/backend/db_connect.js
+++ b/src/backend/db_connect.js
@@ -84,7 +84,7 @@ async function getEvents(){
  */
 function queryAllTags(tags) {
     const pieces = tags.map((t) => `JSON_CONTAINS(tags, '"` + t + `"', '$')`);
-    const piecesAndTogether = containsPieces.join(' AND ');
+    const piecesAndTogether = pieces.join(' AND ');
     return '(' + piecesAndTogether + ')';
 }
 
@@ -97,6 +97,7 @@ function queryAllTags(tags) {
  */
 function queryBetweenDates(start, end) {
     // events where the start and end times are set
+    // TODO: comparisons seem to not be working. need more work
     const event_time_between = `(event_start_time < FROM_UNIXTIME(${start}) AND event_end_time > FROM_UNIXTIME(${end}))`
 
     // TODO: handle events that are marked all day and therefore don't have a start time and end time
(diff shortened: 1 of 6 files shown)


src/backend/db_connect.js
```diff
@@ -76,20 +76,71 @@ async function getEvents(){
...
 }
 
+/**
+ * Construct a part of a query that selects for all tags.
+ *
+ * \param tags  List of string tags
+ * \returns     A string that serves as a SQL selector for all those tags
+ */
+function queryAllTags(tags) {
+    const pieces = tags.map((t) => `JSON_CONTAINS(tags, '"` + t + `"', '$')`);
```
AnthonyS25223, line 86: Very clean sql query. I didn't even know JSON_CONTAINS was a possibility.

src/backend/db_connect.js
```diff
@@ -76,20 +76,71 @@ async function getEvents(){
...
  */
-async function getEventsWithAllTags(tags){
-    // build up a query for all items that have tag1 AND tag2 AND ... AND tagN
-    // the query looks like JSON_CONTAINS(tags, '"tag1"', '$') AND JSON_CONTAINS(tags, '"tag2"', '$') ...
-    const containsPieces = tags.map((t) => `JSON_CONTAINS(tags, '"` + t + `"', '$')`);
-    const containsAllTags = containsPieces.join(' AND ');
-    const query = 'SELECT * FROM events WHERE ' + containsAllTags;
-
-    // actually make that query and return all matching events
+async function getEventsWithTags(tags) {
```
AnthonyS25223, line 114: Just checking if this string set of tags is coming from an external source, and if so, wherever they are called should be a prepared statement just to make sure that we can avoid SQL injections.

src/backend/test/test_api.js
```diff
@@ -0,0 +1,308 @@
+// testing funcs
```
AnthonyS25223, line 1: The test file as a whole looks great! Additionally thank you for figuring out simulated functions those will be super helpful for later testing and things of that nature!

src/backend/db_connect.js
```diff
@@ -76,20 +76,71 @@ async function getEvents(){
...
+
+/**
+ * Query the database for events between two dates, that also have certain tags.
+ *
+ * \param start  Start date in unix time
+ * \param end    End date in unix time
+ * \param tags   List of tags that are required
+ * \returns      A list of all events between those dates that also have every tag
+ */
+async function getEventsBetweenWithTags(start, end, tags) {
```
AnthonyS25223, line 141: This looks great and a very well-named function!

src/backend/db_connect.js
```diff
@@ -76,20 +76,71 @@ async function getEvents(){
...
+}
+
+/**
+ * Query the database for events between two dates.
+ *
+ * \param start  Start date in unix time
+ * \param end    End date in unix time
+ * \returns      A list of all events between those dates
+ */
+async function getEventsBetween(start, end) {
```
AnthonyS25223, line 127: Assuming start and end will be externally sourced, similar to the prior comment, a prepared statement when querying will probably be the safest when implemented. Not necessary for this review, but I just wanted to include that note.

AnthonyS25223 (changes requested):
The code overall looks great! Thank you so much for your hard work and all of the learning you have done to get tests and the API up and running! Just a few notes left on prepared statements but otherwise content with the work!

src/backend/db_connect.js
almondheil: Yes, this is from the URL parameters the user passes. Good catch. How can we avoid SQL injections with something like this?

almondheil:
Adding pool.escape to prevent SQL injection. Checking that still works on the server now.

src/backend/db_connect.js
almondheil: This should be fixed in most recent commit.

src/backend/db_connect.js
almondheil: This has also been escaped. It can't be tested yet (the way I am making the query does not work) so it is possible that these will result in failed requests, but they fix SQL injection now.

AnthonyS25223 (approved):
All of the changes look good to me for addressing the escaping to avoid SQL injections!

almondheil:
Awesome, thank you!

PR #41, opened 2025-04-07
Allow API to run over HTTPS on server side

almondheil:
This pull request changes the API so that it attempts to create an HTTP and an HTTPS server. If not configured it runs them on port 8080 (HTTP) and port 4443 (HTTPS).

Here's a breakdown.

1. In `src/backend/api.js`, I implemented the extra servers. They take environment config on what ports to run on. 
    - The server's `.env` file has been updated to make sure it runs on the right ports, and functionality has been tested manually.
3. In `src/backend/test/test_api.js`, I changed the tests to request through HTTP server specifically. The HTTPS server will not work through testing, since we can't set up the SSL certs there.
4. In `test procedures/api.md`, I changed the API to be run with sudo (to serve on ports 80 and 443, the default ports for HTTP and HTTPS). I also updated the `curl` commands in the tests so they check the correct server address.

### DIFF
--- a/src/backend/api.js
+++ b/src/backend/api.js
@@ -16,8 +16,9 @@ function run() {
   const app = express();
   app.use(express.json());
 
-  // Create a regular HTTP server too. Backup.
-  const http_port = process.env.HTTP_PORT || 80;
+  // Create a regular HTTP server.
+  // Default to 8080 so non-root users can run it.
+  const http_port = process.env.HTTP_PORT || 8080;
   http_server = http.createServer(app).listen(http_port, () => {
     console.log(`http server listening on port  ${http_port}`);
   });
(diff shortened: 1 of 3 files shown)


src/backend/api.js
```diff
@@ -1,28 +1,49 @@
+const db = require('./db_connect.js');
 const express = require('express');
 const fs = require('fs');
-const db = require('./db_connect.js');
+const http = require('http');
+const https = require('https');
```
EthanEdwinHughes, line 5: Looks like you add the http and https imports here.

src/backend/api.js
```diff
@@ -31,17 +52,27 @@ function run() {
  * Stop running the API. Uses the global `server` var set by run()
  */
 function close() {
-  if (server == null) {
-    return;
+  // close http server
+  if (http_server != null) {
```
EthanEdwinHughes, line 56: Closes the http server correctly based on the current state.

src/backend/api.js
```diff
@@ -31,17 +52,27 @@ function run() {
...
 
-  server.close((err) => {
-    if (err) {
-      console.log(`server closed with status ${err}`)
-    } else {
-      console.log('server closed with no error')
-    }
-  });
+  // close https server
+  if (https_server != null) {
```
EthanEdwinHughes, line 67: Closes the https server correctly based on the current state.

EthanEdwinHughes (approved):
Looks good. I appreciate that the code is well documented and understandable. I don't understand why we are switching the server to handle https but I'm sure it good that we can now.

PR #42, opened 2025-04-07
Document current state of API

almondheil:
Add documentation for the current state of the API, and put these on the docs site. Publically available.

Currently the `/events/between/{start}/{end}/` route isn't working, but it does exist and is queryable. I'll try to fix it soon in another PR.

As you can see, most of the changes are the `api-description.json` file. If you want to preview that (to more easily make your comments) you can go to https://editor.swagger.io/ and choose File > Import file... from the top, then select a downloaded copy.

### DIFF
--- a/docs/api-description.json
+++ b/docs/api-description.json
@@ -1,17 +1,8 @@
 {
   "openapi": "3.0.4",
   "info": {
-    "title": "GetGrinnected - OpenAPI 3.0",
-    "description": "This is the API documentation for GetGrinnected. Everything is still work-in-progress, and changes will happen.",
-    "license": {
-      "name": "MIT",
-      "url": "https://opensource.org/license/MIT"
-    },
+    "title": "GetGrinnected API",
     "version": "1.0"
-  },
-  "externalDocs": {
-    "description": "See the other docs",
-    "url": "https://getgrinnected.sites.grinnell.edu/docs/"
   },
   "servers": [
     {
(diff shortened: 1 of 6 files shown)


docs/api-description.json
```diff
@@ -0,0 +1,294 @@
...
+    }
+  ],
+  "paths": {
+    "/events": {
+      "get": {
+        "tags": [
+          "events"
+        ],
+        "summary": "Find all events.",
+        "description": "Find all events stored in the database.",
```
AnthonyS25223, line 33: Would this get me all the events in the database?  Or is this used as an outer-function that is used to find all events with all of the given tags (exclusively must have all tags to be retrieved)

docs/api-description.json
```diff
@@ -0,0 +1,294 @@
...
+        "operationId": "findAllEvents",
+        "parameters": [
+          {
+            "name": "tag",
+            "in": "query",
+            "description": "Tags that events must have all of. Must be quoted in double quotes.",
```
AnthonyS25223, line 39: Could this be worded to something like "Tags: Events must have all of these tags."? Sorry this is just more of a reading choice I feel like this sounds better to me.

docs/api-description.json
```diff
@@ -0,0 +1,294 @@
...
+            "explode": true,
+            "schema": {
+              "type": "string",
+              "example": "2025-04-08"
+            }
+          },
+          {
+            "name": "tag",
+            "in": "query",
+            "description": "Tags that events must have all of. Must be quoted in double quotes.",
```
AnthonyS25223, line 105: The change of description would also occur here if you choose to make that change.

AnthonyS25223 (changes requested):
Overall this looks great I just left a question and a few potential wording shifts (not sure that my suggested is the most legible but it felt better to me). Thank you so much for your hard work!

docs/api-description.json
almondheil: This gets you all the events in the database by default, or you can narrow which tags it filters for. Does that answer your question?

docs/api-description.json
almondheil: Thank you! Working on this.

docs/api-description.json
AnthonyS25223: That's perfect! Thank you so much!

AnthonyS25223 (approved):
These changes look great and thank you so much for explaining and answer my questions!

PR #43, opened 2025-04-07
Fixing database dropping and adding of events for automatic cron job

AnthonyS25223:
We made some changes to our approach to finishing the database calls so that they run and add/drop as expected and also terminate automatically.


src/backend/db_connect.js
```diff
@@ -52,11 +52,11 @@ async function insertEventsFromScrape(){
                 const startTimeISO = new Date(event.StartTimeISO);
                 const endTimeISO = new Date(event.EndTimeISO);
 
-                // Executes the prepared statement with the event values and fills in with defaul values
+                // Executes the prepared statement with the event values and fills in with default values
                 // where the scraper didn't get information.
                 await pool.query(sql, [event.ID, event.Title, event.Description, event.Location,
                     JSON.stringify(orgs), 0, event.Date, event.Time, isAllDay, startTimeISO,
-                    endTimeISO, JSON.stringify(tags), 0, 0, null, 0 ]);
+                    endTimeISO, JSON.stringify(tags), 0, 0, null, 0 ]).then(console.log(`I added event:  ${event.ID}`))
```
almondheil, line 59: Looks like this isn't strictly necessary, but it doesn't hurt to log these things.

update_daily_events.sh
```diff
@@ -5,9 +5,9 @@ npm ci &>> ~/log/update_daily_events.log
 # drop expired events from json, make json of dropped IDs
 npm run drop_file &>> ~/log/update_daily_events.log
 # drop expired events from database
-npm run drop_db &>> ~/log/update_daily_events.log
+timeout 10 npm run drop_db &>> ~/log/update_daily_events.log
 #add new events to JSON
 npm run get_events &>> ~/log/update_daily_events.log
 #use JSON to add new events to database
-npm run add_to_db &>> ~/log/update_daily_events.log
+timeout 10 npm run add_to_db &>> ~/log/update_daily_events.log
```
almondheil, line 12: Like we discussed together: This is on the jank side of things but it works. Seems like the best option for right now.

src/backend/database_add.js
```diff
@@ -1,4 +1,4 @@
 const db = require('./db_connect.js');
 const fs = require("fs");
 
-db.insertEventsFromScrape().then(process.exit(0))
\ No newline at end of file
+db.insertEventsFromScrape()
```
almondheil, line 4: .then() was causing process to end before events were inserted, so it's good we've removed this!

almondheil (approved):
Everything looks good!

Just for the sake of having it on record, we all know that the timeout thing is not the strictly correct way to do this--but it seems like the only thing we can get to work properly.

PR #45, opened 2025-04-08
Implement Profile Page

AnthonyS25223:
This PR includes:
- Implementation of Settings Page
- Implementation of OrgCard
- Implementation of User data Class
- Updates to testing and documentation
- Some removal of things I realized were unnecessary (navController param on mainPage as an example)
This Page will still need (Next sprint or another sprint):
- Implement logic for unfollow orgs
- Implement logic for changing username
- Implement logic for switching accounts
- Implement logic for changing the profile picture
- Implement state save for light and dark mode state on close of app
- A better follow icon than the heart maybe?


michael-paulin (changes requested):
Looks very good overall! There are just a couple things to fix and adding more documentation to the headers of functions and structs wouldn't hurt.

michael-paulin (approved):
All the reviewed part look good. Good job!

PR #52, opened 2025-04-09
Fixing configuration of XCODE.

TimeBudhil:
Followed tutorial.
Turns out, the ViewInspector was incorrectly linked to the main app!
 Unlink it in the project settings (blue icon)


src/ios/GetGrinnected/GetGrinnected/CoreViews/LoginView.swift
```diff
@@ -7,9 +7,6 @@
 
 import Foundation
 import SwiftUI
-import ViewInspector
```
seehorne, line 10: good to remove this! I hadn't previously realized what its inclusion would imply, thanks for putting in the work to figure out how this should work with relation to tests

src/ios/GetGrinnected/GetGrinnected.xcodeproj/project.pbxproj
```diff
@@ -947,11 +926,6 @@
 			package = 0CF63C262DA5E39D009744F8 /* XCRemoteSwiftPackageReference "ViewInspector" */;
 			productName = ViewInspector;
 		};
-		0CF63C462DA5FF20009744F8 /* ViewInspector */ = {
```
seehorne, line 950: thank you for unlinking this from the main app

src/ios/GetGrinnected/GetGrinnected/Components/EventCards.swift
```diff
@@ -131,7 +131,7 @@ struct EventCards: View {
...
-            var myEvents = EventData.parseEvents(json: myjson)
+            let myEvents = EventData.parseEvents(json: myjson)
```
seehorne, line 134: this just gave a warning, but good we fixed it anyway: good practices!

seehorne (approved):
Thank you for putting in the work to fix this Budhil! Looks great, and builds now so yay!

PR #53, opened 2025-04-11
Ellie/scrape pages deeper

seehorne:
DO NOT MERGE THIS UNTIL AFTER DEMO, DON'T WANT TO BREAK CODE FREEZE

anyways, this now scrapes all the events not just the first page of events. makes things like week and month view more viable because we can see more than like 2 days ahead.


src/backend/scrape.js
```diff
@@ -35,25 +36,50 @@ function processExisting(path){
  * Scrapes JSON data of calendar events from specified URL
  * Repackages relevant data to JSON structure stored in event_data.json
  * @param {*} url -> url to data to be scraped, assumes JSON data
+ * @param {*} path -> filepath to write to
  */
-async function scrapeData(url, path) {
```
almondheil, line 39: Can we revert the change on this specific line? I think this breaks style

update_daily_events.sh
```diff
@@ -5,9 +5,9 @@ npm ci &>> ~/log/update_daily_events.log
 # drop expired events from json, make json of dropped IDs
 npm run drop_file &>> ~/log/update_daily_events.log
 # drop expired events from database
-timeout 10 npm run drop_db &>> ~/log/update_daily_events.log
+timeout 20 npm run drop_db &>> ~/log/update_daily_events.log
```
almondheil, line 8: Makes sense that it'll take longer now. It kinda shows the iffyness of this method, but it's the best we've got.

src/backend/scrape.js
```diff
@@ -84,25 +110,13 @@ async function scrapeData(url, path) {
         else if (zeroEventsNotCorrupt){
           stringifyEvent = '\n'+stringifyEvent;
         }
-        appendPromises.push(fs.appendFileSync(path, 
-          stringifyEvent, function(err){
-            if(err) throw err;
-            console.log('WRITING TO JSON')
-          }));
+        fs.appendFileSync(path, stringifyEvent);
+        console.log("Adding "+ event.title + " " + event.id + " " + event.date)
```
almondheil, line 114: Glad to have more log-level info on what we are actually adding!

almondheil (changes requested):
Requesting changes for one weird style thing that seems like it got messed up. The code itself looks sound, and I'll be happy to approve once that line is fixed!

src/backend/scrape.js
seehorne: Done!

PR #54, opened 2025-04-12
Ellie/remove cancelled events

seehorne:
drop events that get cancelled or otherwise removed from Grinnell side prior to their occurence (so we dont have to wait for them to pass by date if they're not actually going to happen)


src/backend/scrape.js
```diff
@@ -133,80 +133,133 @@ async function scrapePage(url, path, anyExistingEvents, zeroEventsNotCorrupt) {
...
+  if (time_based){
+    //record current time for comparison
+    now = new Date(new Date().toLocaleString("en-US", { timeZone: "America/Chicago" }));
+    console.log(now)
+    if (lines.length <= 4){
+      fs.appendFileSync(DROPPATH, CLOSEFILE, function(err){
+        if(err) throw err;
+        console.log('WRITING TO JSON')
+        });//write the end on
+      return; //this implies there are no events even there
```
AnthonyS25223, line 152: What does this doc mean about this "implies there are no events even there?" what is the even there portion meaning?

src/backend/scrape.js
```diff
@@ -133,80 +133,133 @@ async function scrapePage(url, path, anyExistingEvents, zeroEventsNotCorrupt) {
...
+      pageEvents.data.forEach(event => {
+        console.log(currentIDs.size)
+        currentIDs.delete(event.id)
+        console.log("removed "+event.id+ " from set")
+        console.log(currentIDs.size)
+      })
+    }
+    recordedCancelledEvents = currentIDs.size
+    firstAdd = true;
+    if (recordedCancelledEvents){//triggers if set is not empty
```
AnthonyS25223, line 224: This doc is super helpful for legibility!

update_daily_events.sh
```diff
@@ -10,4 +10,8 @@ timeout 20 npm run drop_db &>> ~/log/update_daily_events.log
 npm run get_events &>> ~/log/update_daily_events.log
 #use JSON to add new events to database
 timeout 20 npm run add_to_db &>> ~/log/update_daily_events.log
+#drop any events that are in our up to date db but not in the site itself (cancelled?)
+npm run drop_cancelled  &>> ~/log/update_daily_events.log
```
AnthonyS25223, line 14: I was going to ask how we were handling this on the database side, but this is super helpful in understanding!

AnthonyS25223 (changes requested):
Overall the code looks great and I think following the method of storing ids for cancelled events to drop is great, to avoid needing to also update a db call! Just a little question on a specific part of documentation but otherwise happy with it! Nice job!

AnthonyS25223 (approved):
The changes look great to me thank you for the additional clarity in the documentation!

PR #55, opened 2025-04-12
Restyled header, with text input and searchbar on or off

TimeBudhil:
- every header is within a geometry reader for relative sizing
- adjusted event cards so that they are all same size with this code: .lineLimit(1, reservesSpace: true)


src/ios/GetGrinnected/GetGrinnected.xcodeproj/project.pbxproj
```diff
@@ -719,8 +718,7 @@
 				INFOPLIST_KEY_UIApplicationSceneManifest_Generation = YES;
 				INFOPLIST_KEY_UIApplicationSupportsIndirectInputEvents = YES;
 				INFOPLIST_KEY_UILaunchScreen_Generation = YES;
-				INFOPLIST_KEY_UISupportedInterfaceOrientations_iPad = "UIInterfaceOrientationPortrait UIInterfaceOrientationPortraitUpsideDown UIInterfaceOrientationLandscapeLeft UIInterfaceOrientationLandscapeRight";
-				INFOPLIST_KEY_UISupportedInterfaceOrientations_iPhone = "UIInterfaceOrientationPortrait UIInterfaceOrientationLandscapeLeft UIInterfaceOrientationLandscapeRight";
+				INFOPLIST_KEY_UISupportedInterfaceOrientations = UIInterfaceOrientationPortrait;
```
seehorne, line 721: changes app to always be in portrait. seems like a good call, don't know what rendering landscape would look like for us. maybe we'll get bit with it on the calendar part? but that will probably also be ok

src/ios/GetGrinnected/GetGrinnected/Components/Header.swift
```diff
@@ -7,27 +7,157 @@
...
+func Header(_ safeAreaTop: CGFloat, title: String, searchBarOn: Bool) -> some View {
     
-    var body: some View {
-        NavigationStack{
-            VStack {
-                
-            } //VStack
+    VStack (){
+        //only have this on for the homescreen view
+        if(searchBarOn){
```
seehorne, line 15: this is a really useful feature. do we have that implemented or just UI?

src/ios/GetGrinnected/GetGrinnected/Components/Header.swift
```diff
@@ -7,27 +7,157 @@
...
+                .clipShape(Circle())
+                .background{
+                    Circle()
+                        .fill(.white)
+                        .padding(-2)
+                }
+        }
+    }//HStack for searching
+}
+//Header: View {
```
seehorne, line 74: if this is all commented out anyways, can we possibly remove it? unless you think we'll need to reference it again

src/ios/GetGrinnected/GetGrinnected/Components/Header.swift
```diff
@@ -7,27 +7,157 @@
...
+//            .ignoresSafeArea(.all, edges: .top)
+//        }
+//                
+//    
+//    } //NavigationStack
+//} //body
 
 #Preview {
-    Header(title: "Title")
+    GeometryReader{proxy in
```
seehorne, line 141: Geometry reader was a great find, thank you for looking into that

src/ios/GetGrinnected/GetGrinnected/CoreViews/HomescreenView.swift
```diff
@@ -19,31 +19,44 @@ struct HomescreenView: View {
...
+                    
+                        EventListView()
+
+                    } //VStack
+                    .frame(minHeight: proxy.size.height)//height
+                            
+                }
+            }//Scroll view
+            .edgesIgnoringSafeArea(.top)
+//https://stackoverflow.com/questions/67873845/why-my-custom-views-try-to-take-all-the-available-vertical-space-in-vstack
```
seehorne, line 57: love to see a citation, thank you for crediting your work!

seehorne (changes requested):
This all looks awesome! My one thing is that theres like 60ish commented out lines in Header.swift. Could we remove those? Or if you think there's utility to keeping them there for reference, at least clean up the extraneous whitespace around them. After that I'll be happy to approve

src/ios/GetGrinnected/GetGrinnected.xcodeproj/project.pbxproj
TimeBudhil: Maybe! Thanks for the note about the calendar. that's a good thing to note for later.

src/ios/GetGrinnected/GetGrinnected/Components/Header.swift
TimeBudhil: This is just UI! 
The search bar needs yet to be implemented.

src/ios/GetGrinnected/GetGrinnected/Components/Header.swift
TimeBudhil: Oh yes, that's true! 
I forgot to delete it.

seehorne:
This looks great! Idk why the node is failing but I don’t think it’s bc of this so I’ll put this through then fix it

seehorne (approved):

PR #56, opened 2025-04-12
Ellie/fix node ci

seehorne:
This is off of Budhil's branch so I could see what if anything made it hurt the node, so you only need to actually look at the non-swift files. But there was a bug in a CI test, now I have fixed it


src/backend/test/test_scraper.js
```diff
@@ -79,18 +79,22 @@ test('All events are unique by ID', { concurrency: true }, t => {
       await scrape.dropPastEvents(scrape.CIPATH, true);
       secondScrape = fs.readFileSync(scrape.DROPPATH, 'utf-8');
       secondIDs = JSON.parse(secondScrape);
-      secondIDCount = firstIDs.data.length;
+      secondIDCount = secondIDs.data.length;
```
almondheil, line 82: Ah oops that'll do it

src/backend/test/test_scraper.js
```diff
@@ -79,18 +79,22 @@ test('All events are unique by ID', { concurrency: true }, t => {
...
       //add IDs to set
       firstIDs.data.forEach(event => {
         allIDs.add(event.ID);
+        console.log("adding "+ event.ID + " scrape2")
       });
       secondIDs.data.forEach(event => {
         allIDs.add(event.ID);
+        console.log("adding "+ event.ID + " scrape2")
       });
+      console.log("set size: "+ totalNumberID)
```
almondheil, line 97: Glad to see more logging of what's going on with names on the data

src/ios/GetGrinnected/GetGrinnected.xcodeproj/project.xcworkspace/xcuserdata/ellieseehorn.xcuserdatad/UserInterfaceState.xcuserstate
almondheil, line 1: Not strictly related, but I guess we're just keeping these user-specific xcode states in the repo now. I guess it should be fine.

almondheil (approved):
Looks good! Thanks for fixing this.

PR #57, opened 2025-04-14
Fix DarkTheme and Light Theme Switching

AnthonyS25223:
Making this to have our dark light mode in main so then on a persistent state branch I can interact with it. Generally, this just allows our dark/light mode switch to work as intended but resets when the app is closed due to a non-persistent state.

### DIFF
--- a/src/android/GetGrinnected/MyApplication/app/src/main/java/com/example/myapplication/MainActivity.kt
+++ b/src/android/GetGrinnected/MyApplication/app/src/main/java/com/example/myapplication/MainActivity.kt
@@ -46,6 +46,11 @@ class MainActivity : ComponentActivity() {
             // val eventsTimeFixed = fixTime(eventssorted,length)
             val darkTheme = mutableStateOf(false)
 
+            /* The dark theme value is passed down to the switch that we toggle it at
+            and then the onToggleTheme is a lambda function that allows us to switch
+            the state of dark theme, additionally the darktheme value traces into our
+            theme file which allows us to change the whole app theme.
+            */
             setContent {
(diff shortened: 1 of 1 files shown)


src/android/GetGrinnected/MyApplication/app/src/main/java/com/example/myapplication/MainActivity.kt
```diff
@@ -50,13 +44,13 @@ class MainActivity : ComponentActivity() {
...
+            val darkTheme = mutableStateOf(false)
 
             setContent {
-                MyApplicationTheme {
+                MyApplicationTheme(darkTheme = darkTheme.value) {
                     AppNavigation(
-                        darkTheme = darkTheme,
-                        onToggleTheme = {darkTheme = it},
+                        darkTheme = darkTheme.value,
+                        onToggleTheme = { darkTheme.value = it },
```
michael-paulin, line 53: Should there be spaces on the ends of the bracket contents?

michael-paulin (changes requested):
Looks good! Just think you should put comments explaining what you are doing when setting up the dark mode even if it looks self explanatory.

src/android/GetGrinnected/MyApplication/app/src/main/java/com/example/myapplication/MainActivity.kt
AnthonyS25223: This is just the style of Kotlin generally accepted

michael-paulin (approved):
Great! Thanks for putting in that comment! You can merge now.

PR #59, opened 2025-04-14
Proof-of-concept sending emails

almondheil:
This PR adds a script giving a proof-of-concept that we can send emails using an app password for the Get Grinnected gmail account. Tested from the server, it seems to work--no clue what Ella's been dealing with but we seem okay on Grinnell network.

No tests have been written since this isn't going into production. It's just here so we have access to the code going forward.

### DIFF
--- a/src/backend/send_test_mail.js
+++ b/src/backend/send_test_mail.js
@@ -1,0 +1,30 @@
+let nodemailer = require('nodemailer');
+require('dotenv').config();
+
+let mailTransporter = nodemailer.createTransport(
+    {
+        service: 'gmail',
+        auth: {
+            user: process.env.GMAIL_ACCOUNT,
+            pass: process.env.GMAIL_PASSWORD
+        }
+    }
+);
+
+let mailDetails = {
+    from: process.env.GMAIL_ACCOUNT,
+    to: process.env.FAKE_EMAIL_DELETEME, // TODO: when deleting this code, delete this line in .env
+    subject: "If you're reading this it WORKED!",
+    text: "And I am so proud of my child the auto email maker, even though I did so little work."
+};
+
+mailTransporter.sendMail(
+    mailDetails,
+    function (err, data) {
+        if (err) {
+            console.error(err);
+        } else {
+            console.log('Email sent successfully');
+        }
+    }
+);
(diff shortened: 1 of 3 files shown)


src/backend/send_test_mail.js
```diff
@@ -0,0 +1,30 @@
...
+            pass: process.env.GMAIL_PASSWORD
+        }
+    }
+);
+
+let mailDetails = {
+    from: process.env.GMAIL_ACCOUNT,
+    to: process.env.FAKE_EMAIL_DELETEME, // TODO: when deleting this code, delete this line in .env
+    subject: "If you're reading this it WORKED!",
+    text: "And I am so proud of my child the auto email maker, even though I did so little work."
```
seehorne, line 18: this is cute, obviously we'll change it in the actual thing, but fun touch

src/backend/send_test_mail.js
```diff
@@ -0,0 +1,30 @@
...
+        auth: {
+            user: process.env.GMAIL_ACCOUNT,
+            pass: process.env.GMAIL_PASSWORD
+        }
+    }
+);
+
+let mailDetails = {
+    from: process.env.GMAIL_ACCOUNT,
+    to: process.env.FAKE_EMAIL_DELETEME, // TODO: when deleting this code, delete this line in .env
```
seehorne, line 16: this is where their email would go, correct?

src/backend/send_test_mail.js
```diff
@@ -0,0 +1,30 @@
+let nodemailer = require('nodemailer');
+require('dotenv').config();
+
+let mailTransporter = nodemailer.createTransport(
+    {
+        service: 'gmail',
+        auth: {
+            user: process.env.GMAIL_ACCOUNT,
```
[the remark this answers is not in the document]
almondheil: Yes, this is our email. `process.env.FAKE_EMAIL_DELETEME` is currently my gmail, but it would get switched out for their grinnell email.

seehorne (commented):
This looks solid to me. I'm not yet approving it for merge in bc it appears to mostly be a template but this is great to have on hand

src/backend/send_test_mail.js
almondheil: yes!

seehorne (approved):
I now understang the plan was to have this here until we were ready to have a real one so I'm approving it

PR #60, opened 2025-04-14
Added header to EventTags and edited header of the CheckBox structure.

michael-paulin:


src/ios/GetGrinnected/GetGrinnected/Classes/EventTags.swift
```diff
@@ -7,6 +7,9 @@
 
 import SwiftUI
 
+/*
+ These are the tags an event structure can have. The list is not fully complete.
```
almondheil, line 11: This comment makes me wonder "what is the complete list?" but I know the answer is that we don't have the ability to get a complete list.

It's not a docs concern, but we will want to fix this issue in the future.

src/ios/GetGrinnected/GetGrinnected/Components/CheckBox.swift
```diff
@@ -12,6 +12,8 @@ import SwiftUI
  
  name: A string that holds the displayed name of the option
  isChecked: A boolean that is true if the option is checked. Otherwise, false.
+ uiCompOne: The component that is displayed when the check box is checked.
```
almondheil, line 15: These are good!

It specifies that they are the components displayed, so I am not left wondering what they are used for.

almondheil (approved):
These comments look good! Short of changes to the underlying design (which is for another day) I think this is all set.

PR #63, opened 2025-04-14
Update API so everything should work properly

almondheil:
# Description

A few tweaks to have the API in a good state for whatever we want to do with it going forward--whether that is keeping it in place or deleting it entirely.

These are the major things by file. Hopefully they are helpful to focus on in your review.

api.js:
- tags no longer need to be passed quoted, the API can quote them before passing them to the db
- fix date parsing by going to a date in the format mysql supports (causes changes in db_connect.js)
- make the errors all use the same structure so they are easier to work with

test_api.js:
- update tests so they don't fail, now that things work differently.

db_connect.js:
- change how queries work to match with rest of system

api.md and api-description.json:
- updates to the documentation of the API.

### DIFF
--- a/docs/api-description.json
+++ b/docs/api-description.json
@@ -25,37 +25,12 @@
         "operationId": "findAllEvents",
         "parameters": [
           {
-            "name": "tag",
-            "in": "query",
-            "description": "Events must have all of these tags. Each tag needs double quotes.",
-            "required": false,
-            "explode": true,
-            "schema": {
-              "allOf": [
-                {
-                  "$ref": "#/components/schemas/Tags"
-                }
-              ],
-              "example": [
-                "\"Music\"",
-                "\"Students\""
-              ]
-            }
+            "$ref": "#/components/parameters/tags"
           }
         ],
         "responses": {
           "200": {
-            "description": "successful operation",
-            "content": {
-              "application/json": {
-                "schema": {
-                  "type": "array",
-                  "items": {
-                    "$ref": "#/components/schemas/Event"
-                  }
-                }
-              }
-            }
+            "$ref": "#/components/responses/success"
           }
         }
       }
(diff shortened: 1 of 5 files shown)


almondheil:
Rebased onto main to ensure tests are passing. Probably should have merged instead, but this is okay for this time.

docs/api-description.json
almondheil, line 1: For this file, looking at the whole file in the swagger editor at https://editor.swagger.io/ might be more helpful than looking at the diff.

test procedures/api.md
```diff
@@ -82,59 +82,70 @@ For these steps, open a second terminal not connected to the server.
...
    ```
-   curl 'https://node16049-csc324--spring2025.us.reclaim.cloud/events?tag="doesnotexist"' | jq
+   curl 'https://node16049-csc324--spring2025.us.reclaim.cloud/events?tag=doesnotexist' | jq
    ```
 
 2. Confirm the output is an empty array.
 
 ## Get events between dates
 
-TODO: This section cannot yet be written.
```
almondheil, line 110: This section is actually possible to write and test now.

Of course, I've confirmed that it is working as expected when deploying from my own branch. So I am confident in these instructions.

src/backend/api.js
```diff
@@ -122,45 +118,66 @@ async function getEvents(req, res, next) {
...
+  // that can be handled by the SQL side.
+  // These will look like 'YYYY-MM-DD HH:MM:00.000Z'
+  const startSQLString = new Date(start).toISOString();
+  const endSQLString = new Date(end).toISOString();
+
   // If tags are not requested, simply query db.
   const tags = parseQueryTags(req.query.tag);
-  if (tags == null) { 
-    const events = await db.getEventsBetween(start, end);
+
```
AnthonyS25223, line 170: Very helpful docs for explaining the conditional on what tells us which kind of db query we make

AnthonyS25223 (approved):
Looks great to me I gave one bit of feedback about a word missing in docs, but otherwise nothing but good things to say! Nicely done and that you so much for your hard work!

src/backend/db_connect.js
```diff
@@ -104,12 +104,8 @@ function queryAllTags(tags) {
  * \returns      A string that serves as a SQL selector for those dates
  */
 function queryBetweenDates(start, end) {
-    // events where the start and end times are set
-    // TODO: comparisons seem to not be working. need more work
-    const event_time_between = `(event_start_time < FROM_UNIXTIME(${pool.escape(start)}) AND event_end_time > FROM_UNIXTIME(${pool.escape(end)}))`
-
-    // TODO: handle events that are marked all day and therefore don't have a start time and end time
-    // could it be done when scraping, maybe? this will SUCK to do in mysql.
+    // construct a query for that start and end time
```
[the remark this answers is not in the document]
almondheil: Like we talked about on Discord, this actually revealed an issue in the code that I hadn't tested with the manual test procedure--thank you!

I'm putting this here too so it shows up for accountability/grading purposes.

AnthonyS25223 (approved):
The changes look great thank you so much!

PR #64, opened 2025-04-14
Internal docs lab - almond

almondheil:
# WIP

Not yet ready to be merged. Will be ready when:

- [ ] #63 is merged into main
- [ ] This branch gets updated to match main correctly

# Description

This is my work for the internal documentation lab, so the focus is on documenting code. The diff should show that once the branch is updated to main as described above.

### DIFF
--- a/src/backend/api.js
+++ b/src/backend/api.js
@@ -4,12 +4,15 @@ const fs = require('fs');
 const http = require('http');
 const https = require('https');
 
-/* global var to store if we are running a server */
+/* global var to store the servers we are running,
+ * so they can be shut down when needed */
 var http_server = null;
 var https_server = null;
 
 /**
  * Run the API.
+ * 
+ * This function will not exit on its own, 
  */
 function run() {
   // Load environment vars from .env file
@@ -26,7 +29,7 @@ function run() {
     console.log(`http server listening on port  ${http_port}`);
   });
 
-  // Read SSL credentials from their system files
+  // Read our SSL credentials from the environment (loaded from dotenv)
   const credentials = {
     key: process.env.SSL_KEY,
     ca: process.env.SSL_CA,
@@ -42,14 +45,21 @@ function run() {
       console.log(`https server listening on port ${https_port}`);
     });
 
-  // Define routes
+  // Default route just shows the API is online
   app.get('/', getAPIOnline);
+
+  // Getting all events takes no parameters
   app.get('/events', getEvents);
+
+  // Getting events between, set the URL parameter names with :start and :end.
+  // Those params will get passed into the function as part of `req.params` dictionary
   app.get('/events/between/:start/:end', getEventsBetween);
 }
 
 /**
- * Stop running the API. Uses the global `server` var set by run()
+ * Stop running the API.
+ * 
+ * It will only do something if you called run() before.
  */
 function close() {
   // close http server
@@ -78,8 +88,8 @@ function close() {
 /**
  * Just respond that the API is online.
  *
- * \param req Express request object
- * \param res Express response object
+ * @param req Express request object
+ * @param res Express response object
  */
 function getAPIOnline(req, res) {
   res.send('API online!');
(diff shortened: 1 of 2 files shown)


src/backend/api.js
```diff
@@ -164,86 +175,101 @@ async function getEventsBetween(req, res, _next) {
...
-
+  
   // If there are no tags query the DB normally,
   // if there are tags query for them
+  const tags = parseQueryTags(req.query.tag);
   if (tags.length === 0) {
     const events = await db.getEventsBetween(startSQLString, endSQLString);
     res.json(events);
     return;
+  } else {
```
almondheil, line 186: Decided to change this from an implicit to an explicit else, since that makes more sense when reading it.

This doesn't change the operation of the code.

src/backend/api.js
```diff
@@ -164,86 +175,101 @@ async function getEventsBetween(req, res, _next) {
...
   const validDate = "\\d\\d\\d\\d-\\d\\d-\\d\\d"  // YYYY-MM-DD
   const validTime = "T\\d\\d:\\d\\d"              // THH:MM, no seconds
   const validTimeZone = "[+-]\\d\\d\\d\\d"        // Timezone, eg -0500 or +1230
 
-  // If all features are specified, use that and respect the date and timezone provided
-  // YYYY-MM-DDTHH:MM-0000
-  const allFeaturesRegex = new RegExp('^' + validDate + validTime + validTimeZone + '$');
+  // MATCHES: YYYY-MM-DDTHH:MM-ZZZZ
+  // CHANGES: nothing
+  const allFeaturesRegex = new RegExp(
```
almondheil, line 208: Broken into multiple lines since this is the first instance where a regex is getting combined, and I wanted to explain the parts.

src/backend/api.js
```diff
@@ -164,86 +175,101 @@ async function getEventsBetween(req, res, _next) {
...
+  // If there is only one element,
+  // we need to put it in an array for the next step to work
   if (!Array.isArray(queryTags)) {
     queryTags = [queryTags];
   }
 
-  // Split any comma-separated items like a,b,c to make them part of the array.
-  queryTags = queryTags.map((x) => x.split(',')).flat();
+  // Split any comma-separated items to make the array flat. See example.
+  queryTags = queryTags        // ["a", "b,c"]
```
almondheil, line 267: Split each line so I could show a side-by-side of an example playing out, which explains the code way better

seehorne: This is a good touch! Makes it a lot more clear

src/backend/api.js
```diff
@@ -89,16 +99,18 @@ function getAPIOnline(req, res) {
  * Query the database and respond with all known events in JSON.
  * Supports querying for tags with query parameter `tag`.
  *
- * \param req  Express request object
- * \param res  Express response object
- * \param next Error handler function
+ * @param req  Express request object
```
seehorne, line 102: Looks like javadocs

src/backend/api.js
```diff
@@ -89,16 +99,18 @@ function getAPIOnline(req, res) {
...
-  // Determine if the user wants to query specific tags
+  // Pass any query parameters named "tag" so we can parse them into an array.
+  // Could be undefined, that function will handle it.
   const tags = parseQueryTags(req.query.tag);
 
-  // Query the DB, including tags if those were provided
-  if (tags.length === 0) { 
+  // If there are no tags query the DB normally,
+  // if there are tags query for them
+  if (tags.length === 0) {
```
seehorne, line 113: useful check!

src/backend/api.js
```diff
@@ -164,86 +175,101 @@ async function getEventsBetween(req, res, _next) {
...
  *
- * \param timeParam String representing a date.
- * \returns Unix time equivalent if parseable, NaN if not.
+ * @param timeParam String representing a date.
+ * @returns Unix time equivalent if parseable, NaN if not.
  */
 function parseParamDate(paramDate) {
-  // these are the three features we care about in our date format
+  // these are the three features we care about in our date format.
+  // "\\d" goes to the string \d, which means any digit in regex.
```
seehorne, line 201: This is a helpful explanation, would not have necessarily know what I was looking at otherwise

src/backend/api.js
```diff
@@ -164,86 +175,101 @@ async function getEventsBetween(req, res, _next) {
...
-  // If none of those attempts to figure out the format worked, bypass Date.parse and return failure.
+  // If none of those attempts to figure out the format worked, fail with NaN
+  // (which is the same error that Date.parse returns, making checking easier)
   return NaN;
 }
 
 /**
  * Parse from a query tag object to a well-formed list of individual tags. Tags
  * with commas will be split to support queries like `?tag=a,b,c,d`.
+ *
```
seehorne, line 243: Good to enumerate these!

seehorne (approved):
This is much clearer than previously, thank you for doing this!

PR #65, opened 2025-04-15
Ellie/fix allday events

seehorne:


src/backend/scrape.js
```diff
@@ -96,8 +96,10 @@ async function scrapePage(url, path, anyExistingEvents, zeroEventsNotCorrupt) {
         eventInfo["Time"] = event.date_time;
         eventInfo["StartTimeISO"]= event.date_iso;
         eventInfo["EndTimeISO"]= event.date2_iso;
-        if (eventInfo.StartTimeISO===null){
+        allDayEnd = new Date(event.date2_iso) - new Date('1970-01-01T00:00:00.000Z') === 0;
+        if (eventInfo.EndTimeISO===null || allDayEnd){
```
almondheil, line 100: Good, first we check if the end time is null or if the end time is the unix epoch.

src/backend/scrape.js
```diff
@@ -96,8 +96,10 @@ async function scrapePage(url, path, anyExistingEvents, zeroEventsNotCorrupt) {
...
           eventInfo["AllDay?"] = true;
+          eventInfo["EndTimeISO"] = new Date(new Date(event.date2_iso).setHours(23, 59, 0, 0))
```
almondheil, line 102: Then we set the end time to 23:59:00:00. Yay!

almondheil (approved):
Looks good! I only have two comments to give because there's only a couple little changes, but everything should work! Including with timezones, I was worried about those but it's gonna be okay.

PR #67, opened 2025-04-15
Use `new Date` instead of `Date.parse` in API

almondheil:
This way we aren't throwing around the date in milliseconds and needing to convert it anymore. Which is an improvement.

Updated both code and test files.

### DIFF
--- a/src/backend/api.js
+++ b/src/backend/api.js
@@ -131,14 +131,15 @@ async function getEvents(req, res, _next) {
  */
 async function getEventsBetween(req, res, _next) {
   // Parse the required start and end parameters held in the request
-  const start = parseParamDate(req.params.start);
-  const end = parseParamDate(req.params.end);
+  var start = parseParamDate(req.params.start);
+  var end = parseParamDate(req.params.end);
 
-  // Return descriptive error on failure to parse stard and/or end
+  // Return descriptive error on failure to parse stard and/or end date
+  //
   //   .status(400)  creates a 400 "Invalid request" HTTP error
   //   .json(...)    lets us specify what JSON object we return with the error, 
   //     so we can have a consistent error code and more detailed error messages 
-  if (isNaN(start) && isNaN(end)) {
+  if (isNaN(start.valueOf()) && isNaN(end.valueOf())) {
     res.status(400).json({
       'error': 'Invalid date',
       'message': 'Start and end date could not be read properly.'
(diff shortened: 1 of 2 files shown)


src/backend/api.js
```diff
@@ -228,13 +230,12 @@ function parseParamDate(paramDate) {
...
-    return Date.parse(paramDate);
+    return new Date(paramDate);
   }
 
-  // If none of those attempts to figure out the format worked, fail with NaN
-  // (which is the same error that Date.parse returns, making checking easier)
-  return NaN;
+  // If none of those attempts to figure out the format worked,
+  // return a date that we know is definitely invalid to match the return type.
+  return new Date('purposefully invalid date string');
```
seehorne, line 238: this is kind of silly, but a good solution to a null date being January 1 1970 for some reason

src/backend/test/test_api.js
```diff
@@ -184,65 +184,65 @@ describe('parseParamDate', () => {
...
+        assert.strictEqual(expected.toISOString(), actual.toISOString());
     });
 
     it('should assume Grinnell time if timezone unspecified', () => {
         const input = '1999-01-01T08:19';
-        const expected = Date.parse(input + '-0500');
+        const expected = new Date(input + '-0500');
         const actual = api.parseParamDate(input);
-        assert.strictEqual(expected, actual);
+        assert.strictEqual(expected.toISOString(), actual.toISOString());
```
seehorne, line 203: stringify is a good way to ensure the equals assertions work

src/backend/api.js
```diff
@@ -173,18 +174,18 @@ async function getEventsBetween(req, res, _next) {
...
   // if there are tags query for them
   const tags = parseQueryTags(req.query.tag);
   if (tags.length === 0) {
-    const events = await db.getEventsBetween(startSQLString, endSQLString);
+    const events = await db.getEventsBetween(start, end);
     res.json(events);
     return;
   } else {
-    const events = await db.getEventsBetweenWithTags(startSQLString, endSQLString, tags);
+    const events = await db.getEventsBetweenWithTags(start, end, tags);
```
seehorne, line 188: i like how we are using shorter names for this, which you defined earlier. it is better for style and easy to read

seehorne (approved):
This appears to update date handling on the API side, and also reparameterize some related tests. Looks good to me

PR #74, opened 2025-04-18
Ellie/handle event updates

seehorne:
Now we re-add events every time, so that we can capture when they are not cancelled but do have updates to them. This also involved changing the cron job to do all drops first, since the JSON would be reworked every time an additive scrape happened.


src/backend/scrape.js
```diff
@@ -35,35 +35,30 @@ function processExisting(path) {
  * 
  * Scrapes JSON data of calendar events from specified URL
  * Repackages relevant data to JSON structure stored in event_data.json
+ * Rewrites all events, to ensure catching updates
```
almondheil, line 38: ✅ Gotta keep those docs up-to-date :D

src/backend/scrape.js
```diff
@@ -35,35 +35,30 @@ function processExisting(path) {
...
-    if (!anyExistingEvents){
-      didAdd = lineTracker > 4;
-      anyExistingEvents = didAdd
-      }
-    if (zeroEventsNotCorrupt){
-      stillNone = lineTracker === 4;
-      zeroEventsNotCorrupt = stillNone;
+    if (i==1){
+      //true means adding first event
+      existingIDs = await scrapePage(pageURL, path, existingIDs, true);
```
almondheil, line 57: ✏️ I dug a little and found we do this so it knows it's adding the first event (changed from `zeroEventsNotCorrupt`), so I was wondering if you could add a comment explaining why it needs to know when the first event is happening.

I have an idea of why, but it'd be good to have it documented.

src/backend/scrape.js
```diff
@@ -35,35 +35,30 @@ function processExisting(path) {
...
-    await scrapePage(pageURL, path, anyExistingEvents, zeroEventsNotCorrupt);
-    lineTracker = fs.readFileSync(path, 'utf-8').split('\n').length;
-    if (!anyExistingEvents){
-      didAdd = lineTracker > 4;
-      anyExistingEvents = didAdd
-      }
-    if (zeroEventsNotCorrupt){
-      stillNone = lineTracker === 4;
-      zeroEventsNotCorrupt = stillNone;
+    if (i==1){
```
almondheil, line 55: ❓ I thought `i == 1` meant we were doing the first page of results, not the first result overall. Is that true?

✏️ Also we probably want it to be `===` just for safety's sake

src/backend/test/test_scraper.js
```diff
@@ -138,4 +130,14 @@ test('All events are unique by ID', { concurrency: true }, t => {
         console.log(foundEventCount)
         assert(foundEventCount >= minEvents && foundEventCount <= maxEvents);
       }
-    )
+    );
+
+    test('All events are unique by ID', { concurrency: true }, async t => {
```
almondheil, line 135: ✅ Phew, we're just moving this test. Good and cool

almondheil (changes requested):
I had a couple questions about the scraping logic.

I've also done my best to annotate my comments so they're easier to see at a glance. We have

- ✅ - looks good
- ✏️ - suggested edit
- ❓ - question

almondheil (approved):
Oh, this makes sense. Thank you for explaining with those comments! Approved.

PR #76, opened 2025-04-19
Ellie/ci drop for cancel

seehorne:
This pull requests makes the CI tests for dropping a bit more robust (adds tests for canceled drop equivalent to those we had for timeout drop, and adds tests for both types of drop to check that "removed" events actually get removed)


src/backend/test/test_scraper.js
```diff
@@ -78,66 +78,172 @@ describe('Web Scrape Output Unit Tests', () => {
...
       }
-    )
-    
-    test('After scraping, there are as many or more events as before (no lost events)',
-      { concurrency: true }, async t => {
+      assert(removed === false)
+    }
+    );
+
+  test('When running drop cancel, events actually remove from JSON',
```
michael-paulin, line 111: What does drop cancel mean? Are we testing that if you cancel dropping an event?

src/backend/test/test_scraper.js
```diff
@@ -78,66 +78,172 @@ describe('Web Scrape Output Unit Tests', () => {
...
+        removed = IDSet.delete(remainingEvents.data[i].ID);
+        if (removed== true){
+          break;
+        }
+      }
+      assert(removed === false)
+    }
+    );
+
+  test('When running drop cancel twice, IDs to drop are distinct each run',
```
michael-paulin, line 140: So this checks that none of the events are dropped multiple times?

src/backend/test/test_scraper.js
```diff
@@ -78,66 +78,172 @@ describe('Web Scrape Output Unit Tests', () => {
...
+      beforeSize = ogEvents.data.length;
+      await scrape.dropPastEvents(scrape.CIPATH, false);
+      newEvents = fs.readFileSync(scrape.CIPATH, 'utf-8');
+      nowEvents = JSON.parse(events);
+      nowSize = nowEvents.data.length;
+      assert(nowSize <= beforeSize)
+    }
+  )
+
+  test('After scraping, there are as many or more events as before (no lost events)',
```
michael-paulin, line 201: Could you put more comments on this one?

src/backend/test/test_scraper.js
```diff
@@ -78,66 +78,172 @@ describe('Web Scrape Output Unit Tests', () => {
...
+      //check how many events there are
+      foundEvents = fs.readFileSync(scrape.CIPATH, 'utf-8');
+      jsonifiedFoundEvents = JSON.parse(foundEvents);
+      foundEventCount = jsonifiedFoundEvents.data.length;
+      console.log(foundEventCount)
+      assert(foundEventCount >= minEvents && foundEventCount <= maxEvents);
+    }
+  );
+
+  test('All events are unique by ID', { concurrency: true }, async t => {
```
michael-paulin, line 241: Could you put more comments on this one?

michael-paulin (commented):
Looks good, just a few questions.

src/backend/test/test_scraper.js
seehorne: drop cancel refers to dropping an event because it has been cancelled, as opposed to dropping because it timed out/is over

src/backend/test/test_scraper.js
seehorne: yes, but more directly, every drop run clears the file of dropped IDs because the db immediately takes it to do a drop there, so if we ended up with duplicates we'd end up trying to drop a thing from SQL that SQL no longer has

src/backend/test/test_scraper.js
michael-paulin: Ah ok. Thanks for the clarification.

src/backend/test/test_scraper.js
michael-paulin: Ok that makes sense.

src/backend/test/test_scraper.js
michael-paulin: Good comments.

src/backend/test/test_scraper.js
michael-paulin: Good comments.

michael-paulin (approved):
Looks good! Thanks for explaining the intentions of the tests and adding more comments.

PR #79, opened 2025-04-21
Refactor API into multiple files

almondheil:
This pull request breaks the API into files. Right now there is only one route file, but there will be more with time. It also lets you choose not to start either the HTTP or HTTPS servers depending on your .env file.

It also changes from `.js` to `.cjs` for the extension. This is still the common JS style of doing imports (which works with our testing framework better), but now I am being explicit about the filename.

# Note to reviewer

The diff looks like I wrote a lot of code because moving a file counts as creating it as well, so here is what I have actually changed.

- moved `src/backend/api.js` to `src/backend/api/api.cjs` (within a folder)
- moved all events routes into `src/backend/api/routes/events.cjs`
- these are the main lines I changed in `api.cjs`: [shortlink because URL long](https://github.com/seehorne/GetGrinnected/pull/79/files#diff-bef64ac1fc217dafce70328a9073ff25bb735612bcbcea086dd0a5b624431b1cR26-R55)


src/backend/api/api.cjs
```diff
@@ -0,0 +1,116 @@
...
+ */
+function run() {
+  // Load environment vars from .env file
+  require('dotenv').config();
+
+  // Create the app that we will serve
+  const app = express();
+  app.use(express.json());
+
+  // Read ports for http and https servers. .env loads these values.
```
seehorne, line 26: code changed here

src/backend/api/api.cjs
```diff
@@ -0,0 +1,116 @@
...
+    };
+
+    // Create an HTTPS server serving the application.
+    // Default to 4443 so non-root users can run it.
+    https_server = https
+      .createServer(credentials, app)
+      .listen(https_port, () => {
+        console.log(`https server listening on port ${https_port}`);
+      });
+  }
```
seehorne, line 55: end of code change

src/backend/api/routes/events.cjs
```diff
@@ -0,0 +1,190 @@
+const db = require('../../db_connect.js');
```
seehorne, line 1: locational refactor

src/backend/test/test_api.js
```diff
@@ -185,63 +185,63 @@ describe('parseParamDate', () => {
     it('should accept ISO-8601 time unchanged', () => {
         const input = '2025-04-05T22:19-0500';
         const expected = new Date(input);
-        const actual = api.parseParamDate(input);
+        const actual = events.parseParamDate(input);
```
seehorne, line 188: i see we are changing to use the refactored file here, cool, makes sense

test procedures/api.md
```diff
@@ -30,7 +30,7 @@ To set up the test, do the following steps.
 6. Run the API from your checked-out directory
    ```
    npm ci
-   sudo node src/backend/api.js
+   sudo node src/backend/api/api.cjs
```
seehorne, line 33: we acquired a new folder here, making note of that so we can be aware for future path activities

src/backend/api/routes/events.cjs
seehorne, line 1: It makes sense to move this into a new file because otherwise the api file would get very big, good to split out by type

src/backend/api/api.cjs
```diff
@@ -0,0 +1,116 @@
...
+  // Load environment vars from .env file
+  require('dotenv').config();
+
+  // Create the app that we will serve
+  const app = express();
+  app.use(express.json());
+
+  // Read ports for http and https servers. .env loads these values.
+  // The default ports happen if not specified, and both are chosen because
+  // non-sudo users can create them.
```
seehorne, line 28: having both is good for testing on the server vs testing locally, good call, now we don't have to take it offline to test

src/backend/api/api.cjs
```diff
@@ -0,0 +1,116 @@
...
+
+  // Start up the HTTPS server unless it is disabled by setting port to -1
+  if (https_port !== -1) {
+    // Read our SSL credentials from the environment (loaded by dotenv)
+    const credentials = {
+      key: process.env.SSL_KEY,
+      ca: process.env.SSL_CA,
+      cert: process.env.SSL_CERT,
+    };
+
```
seehorne, line 47: what happens here if neither is true?

seehorne: by which i mean we dont enter either path. is that possible to have happen?

seehorne (approved):

PR #81, opened 2025-04-22
Michael/swift homepage

michael-paulin:
Added date filtering to the home page. Still need to figure out how to update it when viewed day is changed.


src/ios/GetGrinnected/GetGrinnected/Classes/EventListView.swift
```diff
@@ -63,19 +70,56 @@ class EventViewModel: ObservableObject {
...
+     */
+    func filterEventsByDateInterval(timeSpan: DateInterval) {
+        viewedEvents = events.filter { event in
+            // check that the event has a start and end time
+            if event.useful_event_start_time != nil && event.useful_event_end_time != nil {
+                // check current events start or end is in the timeSpan
+                if timeSpan.contains(event.useful_event_start_time!) || timeSpan.contains(event.useful_event_end_time!) {
+                    return true
+                } //if
+            } //if
```
AnthonyS25223, line 102: Commenting on the braces is a good touch to keep track of which they apply to

src/ios/GetGrinnected/GetGrinnected/Components/CheckBox.swift
```diff
@@ -16,7 +16,7 @@ import SwiftUI
  uiCompTwo: The component that is displayed when the check box is not checked.
  */
 struct CheckBoxOption {
-    var name: String
+    var name: String?
```
AnthonyS25223, line 19: Can this name be null? why is that?

src/ios/GetGrinnected/GetGrinnected/Classes/EventListView.swift
```diff
@@ -116,14 +160,17 @@ struct EventListView: View {
                 
         } //For all events
         .onAppear(){
-            viewModel.fetchEvents()
+            viewModel.fetchEvents(timeSpan: timeSpan)
+        }
+        .onChange(of: timeSpan) { oldValue, newValue in
```
AnthonyS25223, line 165: Does this fetch events during a certain time span?

src/ios/GetGrinnected/GetGrinnected/Classes/Event.swift
```diff
@@ -21,7 +23,19 @@ struct Event: Codable {
...
     var event_all_day: Int?
     var event_start_time: String?
+    var useful_event_start_time: Date? {
+        let dateFormatter = DateFormatter()
+        dateFormatter.locale = Locale(identifier: "en_US_POSIX")
+        dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSZ"
+        return dateFormatter.date(from:event_start_time!)!
+    } // referenced from https://stackoverflow.com/questions/36861732/convert-string-to-date-in-swift
     var event_end_time: String?
+    var useful_event_end_time: Date? {
```
AnthonyS25223, line 33: Does this convert the Date time? Could this be made into a function so that start time and end time could just be called as parameters?

AnthonyS25223 (changes requested):
Overall work looks good just had a few questions and wondering if a point of repetition could be avoided with a function. Overall great work!

src/ios/GetGrinnected/GetGrinnected/Components/CheckBox.swift
michael-paulin: We had a problem where if we entered an empty string it would display a dash for the text next to it. I made it optional because the name is simply the text next to the checkbox and we do not always need that.

src/ios/GetGrinnected/GetGrinnected/Classes/EventListView.swift
michael-paulin: Right now it just fetches all the evens, but then it filters then to the time span given.

src/ios/GetGrinnected/GetGrinnected/Classes/Event.swift
michael-paulin: It converts the string into a date object. It could probably get changed into a function later, but right now it is fine how it is.

AnthonyS25223 (approved):

PR #93, opened 2025-04-24
Patch one-time code verification

almondheil:
Very small PR fixing two things

* We had our conditional flipped for verification of one-time codes
* The email we send out needs to include a bit more info.

The changes are very small, so a comment on each change should suffice for review. Thank you!


src/backend/api/routes/user.cjs
```diff
@@ -185,7 +185,7 @@ async function routeVerifyOTP(req, res, _next) {
   //boolean that checks if the code is right
   validCode = await util.otpFileCheck(DBPATH, email, code) 
 
-  if (validCode){
+  if (!validCode){
```
seehorne, line 188: thanks for catching this lol! this should work a lot better

src/backend/one_time_code.cjs
```diff
@@ -52,7 +52,12 @@ function sendCode(email) {
         from: account,
         to: email,
         subject: "Your GetGrinnected One Time Code",
-        text: `Your one-time code is ${code}.`
+        text: `Your one-time code for GetGrinnected is ${code}.
+
```
seehorne, line 56: the additional information is useful context! thanks for adding!

seehorne (approved):
all great catches almond thanks so much!

PR #104, opened 2025-04-25
Fix otp verification async bug by awaiting bcrypt

almondheil:
This PR fixes an OTP bug that Anthony identified.

The solution is just to make sure to `await` when we compare the code, otherwise it returns true by default because a non-evaluated Promise is truthy. Javascript moment.


AnthonyS25223 (approved):
This looks good from what I understand with how promises work in javascript as you mentioned always evaluating to truthy.

PR #105, opened 2025-04-25
Implement JWT verification, create route to get user data from db

almondheil:
This PR implements an express middleware that verifies JWTs, which lets us make sure somebody is logged in. It uses this to create a protected route that lets you get the user data of yourself once you log in and get a token.

# TODO BEFORE MERGE

- [ ] Write internal docs that shows others how to use this
- [x] Update API spec

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,4 +33,5 @@ node_modules
 #
 # Things that should live exclusively server side
 #
-./src/backend/event_data.json
+src/backend/event_data.json
+src/backend/Database/localOTP.db
--- a/docs/api-description.json
+++ b/docs/api-description.json
@@ -764,10 +764,6 @@
             "type": "string",
             "example": "example@grinnell.edu"
           },
-          "password": {
-            "type": "string",
-            "example": null
-          },
           "profile_picture": {
             "type": "string",
             "example": null
--- a/docs/api-description.json
+++ b/docs/api-description.json
@@ -374,6 +374,88 @@
         }
       }
     },
+    "/user/data": {
+      "get": {
+        "tags": [
+          "user"
+        ],
+        "summary": "Get user data when logged in.",
+        "description": "Get user data from the database.\n",
+        "security": [
+          {
+            "auth_token": []
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "Successful request.",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/User"
+                }
+              }
+            }
+          },
+          "401": {
+            "$ref": "#/components/responses/Unauthorized"
+          },
+          "403": {
+            "$ref": "#/components/responses/InvalidToken"
+          },
+          "404": {
+            "$ref": "#/components/responses/NoSuchUser"
+          }
+        }
+      }
+    },
+    "/user/token-refresh": {
+      "post": {
+        "tags": [
+          "user"
+        ],
+        "summary": "Refresh user tokens when logged in.",
+        "description": "Refresh user tokens using a refresh token, which you can only get by logging in previously.",
+        "security": [
+          {
+            "refresh_token": []
+          }
+        ],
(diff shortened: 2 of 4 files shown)


almondheil:
I have decided to ask the team what they would find most helpful by way of internal docs for these features. Because of that, this will now be ready for review.

.gitignore
almondheil, line 1: For reviewer: here I ignore the sqlite db used for one-time passcodes

docs/api-description.json
almondheil, line 1: For reviewer: Here I make changes to the API spec to add the new routes, as well as clarify some other aspects.

src/backend/api/routes/user.cjs
almondheil, line 1: For reviewer: In this file, I implement the route that lets you get your user data, as well as the route that refreshes your tokens once you are logged in.

src/backend/api/api.cjs
almondheil, line 1: For reviewer: This is the main code contribution I have.

It adds a function that verifies a JWT before allowing the next part of a request to run, which we can easily require for any given request.

almondheil (commented):
I've added comments to each file changed explaining what changed.

src/backend/api/routes/user.cjs
```diff
@@ -145,6 +146,30 @@ async function routeSignUpNewUser(req, res, _next) {
...
 
+/**
+ * Generate new user tokens for a specific email address.
+ * 
+ * @param {string} email  Email to generate the tokens for.
+ * @returns  An object with two keys:
+ * - refresh for the user's refresh token
+ * - access for the user's access token
+ */
+function generateUserTokens(email) {
```
AnthonyS25223, line 157: How often should I be refreshing these tokens?

AnthonyS25223 (approved):
Overall looks great just left a few questions I had about some of the logistics of the tokens and calls! Thank you so much for your amazing work it looks great!

docs/api-description.json
```diff
@@ -374,6 +374,88 @@
         }
       }
     },
+    "/user/data": {
+      "get": {
+        "tags": [
+          "user"
+        ],
+        "summary": "Get user data when logged in.",
```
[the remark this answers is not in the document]
almondheil: This uses an auth_token, which is coded by [this specific line](https://github.com/seehorne/GetGrinnected/pull/105/files/f2a893623ab008268e1f98dca2e641f00718c21c#diff-1ca5f87c4555320b279fa628c2b8a40c7ba618b9dc112b4be13178b2dc44270fR386). When they actually get rendered as docs it will show as a lock that if you click, it will tell you what type of token is required.

docs/api-description.json
```diff
@@ -596,6 +749,84 @@
...
+          },
+          "account_name": {
+            "type": "string",
+            "example": "example_user"
+          },
+          "email": {
+            "type": "string",
+            "example": "example@grinnell.edu"
+          },
+          "password": {
```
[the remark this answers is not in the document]
almondheil: Yeah, I can remove them now! That will mean it won't 100% match with the actual returned object until we remove them from the db too, but that's okay.

src/backend/api/routes/user.cjs
almondheil: The authentication token needs to be refreshed every 15 minutes, and you'll know that's necessary if you ever get an HTTP 403 code back from a request. That's the token you'll use to authorize most requests.

The refresh token, on the other hand, only needs to be refreshed before 30 days is up.

The intended workflow for this (which I should write up in a more group-accessible spot) is this.
1. On login or signup (after verifying code), store the tokens you are given.
2. Use the auth code for every request that needs to be authenticated.
3. When the auth code expires, you will start to get 403 errors.
4. Handle those errors by sending a request including your refresh token to the `/user/token-refresh` endpoint, which will give you a fresh auth and refresh token.

PR #106, opened 2025-04-26
Cache Time Out, Refresh Cache on pull down, Refactor UI

AnthonyS25223:
This PR is a little bit of everything:
- Added storing of tokens
- Simplified verification page code (Had some duplication that was unnecessary)
- Add refresh time check to the get for events from API (now refreshes local events every 3hrs)
- Pull down to refresh events with API 
- Change color scheme to match the new blue
- Edit some verification and login/signup error responses to be more legible


src/android/GetGrinnected/MyApplication/app/src/main/java/com/example/myapplication/ApiData.kt
```diff
@@ -23,13 +23,6 @@ data class AuthResponse(
     val access_token: String?
 )
 
-/**
- * Data object of info sent to verify an email is available
- */
-data class EmailRequest(
```
almondheil, line 29: No longer in use anywhere, good to clean up now that it's being replaced with LoginRequest which seems like it is more general.

src/android/GetGrinnected/MyApplication/app/src/main/java/com/example/myapplication/DataStoreSettings.kt
```diff
@@ -21,6 +21,13 @@ object DataStoreSettings {
     private val IS_DARK_MODE = booleanPreferencesKey("is_dark_mode")
     // Preference key for Logged in account id to track the current logged in account
     private val LOGGED_IN_ACCOUNT_ID = intPreferencesKey("logged_in_account_id")
+    // Preference key for Access token
+    private val ACCESS_TOKEN = stringPreferencesKey("access_token")
+    // Preference key for Refresh token
+    private val REFRESH_TOKEN = stringPreferencesKey("refresh_token")
+    // Preference key for the last time we synced with the api
+    private val LAST_SYNC_TIME = longPreferencesKey("last_sync_time")
```
almondheil, line 29: The data store settings file seems like a good place to keep all of these! I know there's a lot of boilerplate for getters and setters, but it looks good.

I also appreciate the ALL_CAPS_NAMES for them.

src/android/GetGrinnected/MyApplication/app/src/main/java/com/example/myapplication/MainActivity.kt
```diff
@@ -31,7 +31,19 @@ class MainActivity : ComponentActivity() {
...
         lifecycleScope.launch {
-            AppRepository.syncFromApi()
+            // Gets the last time we synced with the API
+            val lastSyncTime = DataStoreSettings.getLastSyncTime(applicationContext).first()
+            // Gets the current time
+            val now = System.currentTimeMillis()
+
+            // If the last time we Synced is null ie we have never synced or it has been
+            // More than 3 hrs (10,800,000 milliseconds) since our last sync
+            if (lastSyncTime == null || now - lastSyncTime > 10800000) {
```
almondheil, line 41: Do you think it would be good to change this to `3 * 60 * 60 * 1000`? I think it might make it clearer in addition to the comment, but I'd leave it to your discretion. Just wanted to bring it up.

src/android/GetGrinnected/MyApplication/app/src/main/java/com/example/myapplication/ui/theme/Color.kt
```diff
@@ -2,10 +2,12 @@ package com.example.myapplication.ui.theme
...
 
-val Purple80 = Color(0xFFD0BCFF)
-val PurpleGrey80 = Color(0xFFCCC2DC)
-val Pink80 = Color(0xFFEFB8C8)
-
-val Purple40 = Color(0xFF6650a4)
-val PurpleGrey40 = Color(0xFF625b71)
-val Pink40 = Color(0xFF7D5260)
\ No newline at end of file
+// Colors Tried to
```
almondheil, line 5: It seems like this comment might be unfinished. What are the colors trying to do?

src/android/GetGrinnected/MyApplication/app/src/main/java/screens/EmailVerificationScreen.kt
```diff
@@ -175,15 +180,22 @@ fun EmailVerificationScreen(email: String, flag: Boolean, username: String, navC
                                     context,
                                     email.hashCode()
                                 )
+                                }
+                                // Sets storage preference logged in to true
+                                DataStoreSettings.setLoggedIn(context, true)
+                                // Sets our access token to the token we obtained
+                                DataStoreSettings.setAccessToken(context, response.body()?.access_token!!)
+                                // Sets our refresh token to the token we obtained
+                                DataStoreSettings.setRefreshToken(context, response.body()?.refresh_token!!)
```
almondheil, line 189: To my understanding, there are cases where `?` and `!` can fail--but I don't really know this aspect of the language past that.

Are we already checking for that?

almondheil (changes requested):
I'm requesting changes for this to make sure I can see your response to a couple of my questions.

I'm sorry to slow down the dev process on this one, but I'm hoping the things I asked about aren't going to need changes--I just wanna make sure.

src/android/GetGrinnected/MyApplication/app/src/main/java/com/example/myapplication/MainActivity.kt
AnthonyS25223: This is a fantastic suggestion that would make it so much more clear!

src/android/GetGrinnected/MyApplication/app/src/main/java/com/example/myapplication/ui/theme/Color.kt
AnthonyS25223: Totally forgot about it mid-thought will finish it

src/android/GetGrinnected/MyApplication/app/src/main/java/screens/EmailVerificationScreen.kt
AnthonyS25223: We should be safe using these calls (1. Android Studio's IDE makes a ton of suggestions to ensure your code is safe 2. We are only asserting these values exist when a user has been verified and we have checked that the access_token is not null). So I think this part is okay.

almondheil (approved):
Looks great! Thank you for doing this :)

PR #107, opened 2025-04-26
Change the JSON API description to a YAML one

almondheil:
# Description

This PR changes the API description from a JSON file to a YAML one.

**It does not make any other changes.** I know that literally every line is changed, but I didn't try and sneak in any content changes. This is just an attempt to make this better in the future.

# Rationale

Here's why: The online editor I've been using to generate the file (https://editor.swagger.io) has decided it no longer wants to let me export my file unless I pay them money.

That editor works in YAML though, so if I copy+paste the text out it'll work just fine. It's just a question of whether I need to convert the file manually.

It also may be easier for us to read, which is a good benefit.


docs/api-v1.md
```diff
@@ -1,3 +1,3 @@
 # API v1
 
-!!swagger api-description.json!!
```
AnthonyS25223, line 3: It appears that we are switching file formats as you mentioned in the PR Channel

docs/api-v1.yaml
```diff
@@ -0,0 +1,594 @@
...
+        message:
+          example: Start date could not be read properly.
+    ErrorDateInvalid:
+      allOf:
+        - $ref: '#/components/schemas/Error'
+      properties:
+        error:
+          example: Bad date order
+        message:
+          example: Start date must occur before end date.
```
AnthonyS25223, line 594: Overall the structure looks great thank you for your amazing work!

AnthonyS25223 (approved):
Overall just appears we are switching file formats and not much else so feel like a couple comments is fair! It looks great and thank you so so much for your hard work!

PR #111, opened 2025-04-26
Switch AuthToken Secret to Access Token

AnthonyS25223:
I changed the auth token secret to an access token secret instead.


almondheil (approved):
Yep, it's right now. Thank you.

PR #116, opened 2025-04-27
Allow modification of user data

almondheil:
This PR creates getter and setter routes for three parts of the user data.
- username
- favorited_events
- notified_events

I have tested them a decent amount on the server, but no automated tests still. That's a major hurdle.

It also updates the API docs to reflect how those methods work.


docs/api-v1.yaml
```diff
@@ -291,6 +269,215 @@ paths:
...
+                type: object
+                properties:
+                  favorited_events:
+                    type: array
+                    example: [14, 20, 77]
+                    items:
+                      type: integer
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '403':
```
AnthonyS25223, line 295: Will not this error so that I look for it in some sort of call to ensure my Tokens are refreshed

src/backend/api/utils.cjs
```diff
@@ -367,14 +367,103 @@ function parseQueryTags(queryTags) {
...
+
+/**
+ * Given an object, determine whether it is an array containing valid event IDs.
+ * 
+ * @param {*} object  Object to check.
+ * @returns `true` if these conditions are true.
+ * - `object` is an array
+ * - all elements of `object` correspond to a valid event ID
+ */
+async function eventsExist(object) {
```
AnthonyS25223, line 391: This seems like a super helpful function to have to ensure the event is there!

docs/api-v1.yaml
```diff
@@ -291,6 +269,215 @@ paths:
...
+                    example: "Bad request"
+                  message:
+                    example: "Request body must contain notified_events."
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '403':
+          $ref: '#/components/responses/InvalidToken'
+        '404':
+          $ref: '#/components/responses/NoSuchUser'
+  /user/username:
```
AnthonyS25223, line 422: These docs are great to get an idea of what we need to do to make proper calls!

AnthonyS25223 (approved):
Honestly no feedback other than it looks great and thank you so much for your hard work this is really great!

PR #121, opened 2025-04-28
Check if body parameters exist more rigorously

almondheil:
This PR makes sure we check the body parameters of requests more rigorously.

It does this by creating a middleware (basically a function that verifies a request before it runs) that makes sure the request body contains a key more rigorously--both checking for the key and making sure the request body itself exists. (the difference between a body of `{}` and `undefined`).

This is `middlewareBodyExists`. When you call that, it creates a middleware that verifies that thing.

I also did a simular change to `middlewareVerifyJWT`, to make it take a parameter and construct a new middleware function.

### DIFF
--- a/src/backend/api/api.cjs
+++ b/src/backend/api/api.cjs
@@ -21,6 +21,8 @@ var https_server = null;
 function run() {
   // Load environment vars from .env file
   require('dotenv').config();
+  const ACCESS_JWT = process.env.ACCESS_TOKEN_SECRET
+  const REFRESH_JWT = process.env.REFRESH_TOKEN_SECRET
 
   // Create the app that we will serve
   const app = express();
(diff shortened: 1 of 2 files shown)


src/backend/api/api.cjs
```diff
@@ -179,77 +244,99 @@ function routeShowOnline(_req, res) {
...
+function middlewareVerifyJWT(token_secret) {
+  // Construct a request function that will check that token_secret.
+  // Based on https://www.slingacademy.com/article/authentication-authorization-expressjs-jwt/.
+  return async (req, res, next) => {
+    // Get the contents of the Authorization header sent along with the request.
+    const authHeader = req.get('Authorization');
+
+    // Attempt to get the token out of the header. Two important details:
+    // 
+    // - authHeader will be undefined if it was not included, so we need to use
```
AnthonyS25223, line 290: Very explicit and detailed documentation nicely done

src/backend/api/routes/user.cjs
```diff
@@ -29,15 +29,8 @@ async function routeGetUserData(req, res, _next) {
  * @param {*} _next  Express error handler for async, unused.
  */
 async function routeSendOTP(req, res, _next) {
-  // Make sure email is included in the body
+  // Get email from the body, we know it is there thanks to the middleware
```
AnthonyS25223, line 32: Good point as the the assumptions we can make due to prior function calls

AnthonyS25223 (approved):
Very very clean and well documented code! It looks great! Thank you so much for your hard work and I really only left one comment that may or may not be referencing a typo I am not sure so just wanted to add it just in case!

src/backend/api/api.cjs
```diff
@@ -64,22 +66,23 @@ function run() {
...
+   *****************************************************************************
+   *
+   * Each route has 4 parts. They look like this.
    *
-   * These functions are basically aliases for middlewareVerifyJWT, but each
-   * of them has it verify using a different secret. This lets us verify either
-   * an auth key or a refresh key.
+   * `app.put('a', [b, c], d);`
+   * - put means that this is a PUT request, rather than a GET or something else
+   * - 'a' is the URI, it's the end of the URL to access that route--what frontends connect to.
```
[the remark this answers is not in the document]
almondheil: They're different things, but I didn't get the distinction and was being confused. This thing is actually a URL still, so I'm gonna fix that.

PR #122, opened 2025-04-29
Implement Remaining API Connections

AnthonyS25223:
DONE:
- Setup the routes to make the put requests for favorited events, notified events, and updated username
- I implemented a utility function to make safe API calls specifically regarding an expired access token
- Implemented a query to delete expired events that also removes them from our users list of favorited events
- Changed Event Cards to use the event is notification state instead of the mutable state to ensure it updates properly when we rebuild the app for persistence.
- Launch a background task to update the remote with new local info (currently set to once a minute, can be slowed down)
- Settings page now updates the remote username
- Pull to refresh and sync local information with the remote database.
- Commented out switch account icon

### DIFF
--- a/src/android/GetGrinnected/MyApplication/app/src/main/java/com/example/myapplication/ApiModel.kt
+++ b/src/android/GetGrinnected/MyApplication/app/src/main/java/com/example/myapplication/ApiModel.kt
@@ -5,6 +5,7 @@ import retrofit2.http.Body
 import retrofit2.http.GET
 import retrofit2.http.Header
 import retrofit2.http.POST
+import retrofit2.http.PUT
 
 
 /**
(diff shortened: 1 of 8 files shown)


src/android/GetGrinnected/MyApplication/app/src/main/java/com/example/myapplication/ApiModel.kt
```diff
@@ -34,4 +35,16 @@ interface ApiModel {
     // Async function to refresh auth tokens
     @POST("user/token-refresh")
     suspend fun refreshToken(@Header("Authorization") refreshToken: String): Response<TokenRefreshResponse>
+
```
seehorne, line 38: good to have these as backups of local data, especially if a user deletes and redownloads or uses multiple devices

src/android/GetGrinnected/MyApplication/app/src/main/java/com/example/myapplication/ApiUtils.kt
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Used to safely attempt to make an authorized api call as it handles checking the access token is
+ * valid and if not makes a refresh call for our tokens and then reattempts the api call with the
+ * newly refreshed access token
+ * @param context the current context of the app
+ * @param apiCall the api call you are trying to make
+ * @return the response of the api call
+ */
+suspend fun <T> safeApiCall(context: Context, apiCall: suspend (accessToken: String) -> Response<T>): Response<T> {
```
seehorne, line 20: this is smart, helps the calls generalize

src/android/GetGrinnected/MyApplication/app/src/main/java/com/example/myapplication/AppRepository.kt
```diff
@@ -145,14 +178,93 @@ object AppRepository {
         dao.upsertAccount(account)
 
     }
+
```
seehorne, line 181: these are really useful additions for keeping everything neat, ill likely reference it in the swift implementation

src/android/GetGrinnected/MyApplication/app/src/main/java/screens/SettingsScreen.kt
```diff
@@ -112,7 +112,7 @@ fun SettingsScreen(modifier: Modifier = Modifier,
 
             Spacer(modifier = modifier.weight(1f))
 
-            // This is the icon button for switch an account
```
seehorne, line 115: we no longer need this so thank you for commenting it out

seehorne (approved):
Looks good, some very important auth stuff here, and keeping everything local up to date in the server too is an awesome call

PR #127, opened 2025-04-30
Ellie/make apple demo

seehorne:
This allows for the creation of a demo account for apple to test with so we can get approved for testflight. It redirects the code to a gmail I made if the provided input email we'll have them demo with, so they can still do all the login stuff without having a grinnell email.

### DIFF
--- a/src/backend/api/utils.cjs
+++ b/src/backend/api/utils.cjs
@@ -81,8 +81,12 @@ function validateUsername(username) {
 }
 
 async function sendOTP(email) {
+    sendTo = email
+    if (email == 'getgrinnected.demo@grinnell.edu'){
+        sendTo = 'getgrinnected.demo@gmail.com'
+    }
     // Send the user a one-time code by email.
-    const code = sendCode(email);
+    const code = sendCode(sendTo);
 
     // Calculate the expiration time of the OTP from right now.
     const expiration_mins = 15;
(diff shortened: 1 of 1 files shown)


src/backend/api/utils.cjs
```diff
@@ -81,8 +81,12 @@ function validateUsername(username) {
 }
 
 async function sendOTP(email) {
+    sendTo = email
+    if (email.toLowerCase() == 'getgrinnected.demo@grinnell.edu'){
+        sendTo = 'getgrinnected.demo@gmail.com'
+    }
     // Send the user a one-time code by email.
-    const code = sendCode(email);
+    const code = sendCode(sendTo);
```
TimeBudhil, line 89: yes instead of sending code to email, send it to sendto, that makes sense

src/backend/api/utils.cjs
```diff
@@ -81,8 +81,12 @@ function validateUsername(username) {
 }
 
 async function sendOTP(email) {
+    sendTo = email
+    if (email.toLowerCase() == 'getgrinnected.demo@grinnell.edu'){
+        sendTo = 'getgrinnected.demo@gmail.com'
+    }
```
TimeBudhil, line 87: is it okay that the demo email isn't a real email?

seehorne: since we are never actually sending it to that email, yes. the gmail is a real email, I just made it.

PR #129, opened 2025-05-01
Michael/swift settings page

michael-paulin:
Added a logout button and a text size slider.


src/ios/GetGrinnected/GetGrinnected.xcodeproj/project.pbxproj
```diff
@@ -26,7 +26,6 @@
 		4C02ED382DB68CF40023BCF5 /* DateExtension.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4C02ED372DB68CF40023BCF5 /* DateExtension.swift */; };
 		4CA0C5752DA4BF6500059803 /* CheckBox.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4CA0C5742DA4BF6500059803 /* CheckBox.swift */; };
 		4CA0C5772DA5D16200059803 /* Header.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4CA0C5762DA5D15300059803 /* Header.swift */; };
-		4CCCD6CB2DA5DBC000F1A806 /* Organizations.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4CCCD6CA2DA5DBC000F1A806 /* Organizations.swift */; };
```
TimeBudhil, line 29: fair! We do not need the organizations.swift file

src/ios/GetGrinnected/GetGrinnected/Components/InputField.swift
```diff
@@ -19,11 +19,11 @@ struct InputView: View {
                 .font(.footnote)
             if isSecureField {
                 SecureField(placeholder, text: $text)
-                    .font(.system (size: 14))
+                    .font(.callout)
                     .autocapitalization(.none)
             } else {
                 TextField(placeholder, text: $text)
-                    .font(.system (size: 14))
+                    .font(.callout)
```
TimeBudhil, line 26: I see! This is how it's used..

src/ios/GetGrinnected/GetGrinnected/CoreViews/ContentView.swift
```diff
@@ -11,15 +11,21 @@ import SwiftUI
 struct ContentView: View {
     @AppStorage("isLoggedIn") private var isLoggedIn = false
     
+    // store standard font size of 20 in app storage
+    @AppStorage("FontSize") private var fontSize: Double = 20.0
```
TimeBudhil, line 15: Good use of @AppStorage! That makes complete sense

src/ios/GetGrinnected/GetGrinnected/CoreViews/ContentView.swift
```diff
@@ -11,15 +11,21 @@ import SwiftUI
...
+
```
TimeBudhil, line 16: This is just a me thing, idk what your standards are, but I think variables that have similar function, like SortType, FilterType, and filter, should be next to eachother (in EventList). In this case,  the font and fontSize I would put together. 
If you have some reason for keeping them separate, then this is fine the way it is, but otherwise would you be able to group them together? (Not like I'm doing that all the time either, but I think from hereon out it'd be nice to do that on my side as well)

src/ios/GetGrinnected/GetGrinnected/CoreViews/MainNavView.swift
```diff
@@ -9,6 +9,9 @@ import Foundation
 import SwiftUI
 
 struct MainNavView: View {
+    // boolean to keep track if we are logged in
+    @Binding var isLoggedIn: Bool
```
TimeBudhil, line 13: Ah! To be able to utilize this in the settings page, that makes sense.

src/ios/GetGrinnected/GetGrinnected/CoreViews/SettingsView.swift
```diff
@@ -9,12 +9,19 @@ import Foundation
...
+    
     // access the core data to see the color scheme the device is set to
     @Environment(\.colorScheme) private var userColorScheme
     
+    // boolean to keep track if we are logged in
+    @Binding var isLoggedIn: Bool
     // our users username
     @State private var username: String = "user123"
     
+    @StateObject private var userProfile = UserProfile()
```
TimeBudhil, line 23: Just thinking ahead, no need for any changes, but I would imagine that eventually this will be passed into settingsView? What do you think?

src/ios/GetGrinnected/GetGrinnected/CoreViews/SettingsView.swift
```diff
@@ -30,23 +37,57 @@ struct SettingsView: View {
...
+                        Slider(
+                            value: $fontSize,
+                            in: 20...45,
+                            step: 5
+                        ) {} minimumValueLabel: {
+                            Text("A")
+                                .font(.system(size: 20))
+                        } maximumValueLabel: {
+                            Text("A")
+                                .font(.system(size: 45))
```
TimeBudhil, line 81: I like the use of a slider for text size! I assume the "step:5" allows for the various levels to be sticky/discrete as opposed to continuous? If so, would you mind commenting here about the function?

TimeBudhil (changes requested):
Overall very solid changes! Thanks for cleaning up by deleting those other two files. 
A few small changes (2 or so commenting requests, and a couple questions in regarding how to implement them in the future), but other than that looks great!

TimeBudhil (approved):

src/ios/GetGrinnected/GetGrinnected/CoreViews/ContentView.swift
TimeBudhil: Oh ok! Thanks for noting that.

src/ios/GetGrinnected/GetGrinnected/CoreViews/SettingsView.swift
TimeBudhil: Makes sense!

PR #131, opened 2025-05-01
Refactor user login routes under "session" grouping

almondheil:
This PR changes the routes that deal with the user session from being `/user/...` routes to `/session/...` routes instead.

After this change is deployed, frontends will need to change the URLs they access. It will be simple find-and-replace.

- `/user/login` -> `/session/login`
- `/user/signup` -> `/session/signup`
- `/user/resend-otp` -> `/session/resend-code`
- `/user/verify` -> `/session/verify`
- `/user/token-refresh` -> `/session/refresh`

I will also post an announcement when that does happen.

### DIFF
--- a/docs/api-v1.yaml
+++ b/docs/api-v1.yaml
@@ -6,15 +6,17 @@ servers:
   - url: https://node16049-csc324--spring2025.us.reclaim.cloud/
 tags:
   - name: events
-    description: Getting calendar events. 
+    description: Retrieve calendar events. 
+  - name: session
+    description: Session login and authorization.
   - name: user
-    description: Logging in and changing user settings.
+    description: Get and modify logged-in user settings.
 paths:
-  /user/login:
+  /session/login:
     post:
       tags:
-        - user
-      summary: Log in.
+        - session
+      summary: Log into existing account.
       description: |
         Log in an existing user, which will send a one-time code to their email. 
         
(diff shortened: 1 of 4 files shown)


docs/api-v1.yaml
```diff
@@ -6,15 +6,17 @@ servers:
   - url: https://node16049-csc324--spring2025.us.reclaim.cloud/
 tags:
   - name: events
-    description: Getting calendar events. 
+    description: Retrieve calendar events. 
+  - name: session
+    description: Session login and authorization.
   - name: user
-    description: Logging in and changing user settings.
+    description: Get and modify logged-in user settings.
```
AnthonyS25223, line 13: Big fan of this wording choice

docs/api-v1.yaml
```diff
@@ -6,15 +6,17 @@ servers:
...
 paths:
-  /user/login:
+  /session/login:
     post:
       tags:
-        - user
-      summary: Log in.
+        - session
+      summary: Log into existing account.
```
AnthonyS25223, line 19: Very good discerning note

docs/api-v1.yaml
```diff
@@ -240,11 +242,11 @@ paths:
...
         '404':
           $ref: '#/components/responses/NoSuchUser'
-  /user/token-refresh:
+  /session/refresh:
     post:
       tags:
-        - user
-      summary: Refresh user tokens when logged in.
+        - session
+      summary: Refresh tokens when already logged in.
```
AnthonyS25223, line 249: Good note of the necessity of being logged in to make this call

AnthonyS25223 (approved):
Overall this refactoring makes a lot more sense I think intuitively for thinking about the calls we make! Nicely done thank you so much for your amazing work!

PR #132, opened 2025-05-01
Visual changes

TimeBudhil:
* sortorder name changed from eventTime to time and eventName to name (so sort function displays "name" instead of "eventName" etc.)
* changed default font size to 15 (back to normal size of the search boxes, etc.
* added boxs to all settings options
* adding padding on top of homeview so that the search bar and home view don't overlap
* spellcheck on some comments (weekview)
* adding some padding to search bar, so visually isn't on left-most side of screen
* added explicit call to white color to header text, so settings page no longer looks blue (as it was for some weird reason)
* changed white color to actual white
* removed then added back search bar, as search bar was for some reason not being detected by project

### DIFF
--- a/src/ios/GetGrinnected/GetGrinnected/Components/Header.swift
+++ b/src/ios/GetGrinnected/GetGrinnected/Components/Header.swift
@@ -36,6 +36,7 @@ struct Header: View{
                 Text(title)
                     .font(.title)
                     .bold()
+                    .foregroundStyle(.colorWhite)
             }//hstack for content and buttons
             .padding(.bottom)
             .frame(maxWidth: .infinity)
--- a/src/ios/GetGrinnected/GetGrinnected/CoreViews/HomeView.swift
+++ b/src/ios/GetGrinnected/GetGrinnected/CoreViews/HomeView.swift
@@ -17,7 +17,7 @@ struct HomeView: View {
     
     
     //Sorting and Filtering parameters
-    @State private var sortOrder = SortOrder.eventTime
+    @State private var sortOrder = SortOrder.time
     @State private var filterType = FilterType.name
     @State private var filter = ""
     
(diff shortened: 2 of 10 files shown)


src/ios/GetGrinnected/GetGrinnected/CoreViews/HomeView.swift
```diff
@@ -36,20 +36,20 @@ struct HomeView: View {
                         VStack(spacing: 16) {
                             WeekView(selectedDate: $viewModel.viewedDate)
                                 .padding(.bottom, 4)
-                                .padding(.top, 120)
+                                .padding(.top, 140)
```
michael-paulin, line 39: We probably should not hardcode padding size because it will look bad at different text sizes, but this can be changed later.

src/ios/GetGrinnected/GetGrinnected/CoreViews/SettingsView.swift
```diff
@@ -62,26 +62,58 @@ struct SettingsView: View {
                         // change username field
                         InputView(text: $username, title: "Change Username", placeholder: "Username")
                             .padding()
+                            .background( //background color as app container
```
michael-paulin, line 65: This box looks really nice.

src/ios/GetGrinnected/GetGrinnected/Components/Header.swift
```diff
@@ -36,6 +36,7 @@ struct Header: View{
                 Text(title)
                     .font(.title)
                     .bold()
+                    .foregroundStyle(.colorWhite)
```
michael-paulin, line 39: Thanks for making this white, it was hard to read.

michael-paulin (approved):
Looks really good. The only visual change I think we should add later is to make sure the home view had a header like the other views.

src/ios/GetGrinnected/GetGrinnected/CoreViews/HomeView.swift
TimeBudhil: Yeahh I would agree! I think we use something with geometryreader like in homeview

src/ios/GetGrinnected/GetGrinnected/CoreViews/SettingsView.swift
TimeBudhil: Thanks! same as event cards :)

src/ios/GetGrinnected/GetGrinnected/Components/Header.swift
TimeBudhil: yess I agree!

PR #133, opened 2025-05-01
Implement, test, and document account deletion

almondheil:
This PR adds a route to delete your account by sending a DELETE request to /user.

It also updates the docs for the API to mention that route, and adds tests since it's possible to automatically test the route.

### DIFF
--- a/docs/api-v1.yaml
+++ b/docs/api-v1.yaml
@@ -10,7 +10,7 @@ tags:
   - name: session
     description: Session login and authorization.
   - name: user
-    description: Get and modify logged-in user settings.
+    description: Get and modify logged-in user data.
 paths:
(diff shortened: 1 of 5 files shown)


docs/api-v1.yaml
```diff
@@ -784,7 +809,7 @@ components:
           example: Unknown error
         message:
           type: string
-          example: An unknown error has occured.
+          example: An unknown error has occurred.
```
EthanEdwinHughes, line 812: I like the spelling correction. Looks like a very scary error. Hopefully is very rare.

EthanEdwinHughes (approved):
Looks good. I have nothing to improve. Everything was well documented and clear.

PR #136, opened 2025-05-04
Budhil/links

TimeBudhil:
Links on event cards!


src/ios/GetGrinnected/GetGrinnected/Components/EventCard.swift
```diff
@@ -50,8 +50,24 @@ struct EventCard: View {
     let isExpanded: Bool//The single card does not need a @binding or @state tag
     //simply based on this value, the expansion will show and not.
     
+    let description: String //description used later
+    //let deepLink: URL //for now unneeded
     
     /** Some helper functions for the date*/
```
AnthonyS25223, line 56: Maybe a little more descriptive comment on what this helper function does

src/ios/GetGrinnected/GetGrinnected/Database/UserProfile.swift
```diff
@@ -80,7 +80,7 @@ class UserProfile: ObservableObject {
     }
 
     func loginUser(email: String, completion: @escaping (Result<String, Error>) -> Void) {
-        let url = URL(string: "https://node16049-csc324--spring2025.us.reclaim.cloud/user/login")!
+        let url = URL(string: "https://node16049-csc324--spring2025.us.reclaim.cloud/session/login")!
```
AnthonyS25223, line 83: Nice refactoring of api calls to align with the new structure

src/ios/GetGrinnected/GetGrinnected/Components/EventCard.swift
```diff
@@ -152,6 +168,17 @@ struct EventCard: View {
                         
                         
                         VStack {
+                            ShareLink(
+                                item: "Check out this event!\n",
+                                subject: Text("Join me at this event!"),
+                                message: Text("\(self.description)"),
```
AnthonyS25223, line 174: This looks very simple I wonder if there is a similar way to do this in kotlin as it looks so smooth and clean!

AnthonyS25223 (changes requested):
Overall looks great just had a little question on some documentation but otherwise very happy with it! Keep up the great work!

src/ios/GetGrinnected/GetGrinnected/Components/EventCard.swift
TimeBudhil: Oh actually, this commment is not needed because we no longer have those functions

src/ios/GetGrinnected/GetGrinnected/Database/UserProfile.swift
TimeBudhil: thanks!

src/ios/GetGrinnected/GetGrinnected/Components/EventCard.swift
TimeBudhil: Yes I wonder too!

AnthonyS25223 (approved):
This looks great! Nicely done and thanks for the clarity!

PR #138, opened 2025-05-05
Tweak PUT-ing arrays to ignore invalid items.

almondheil:
In this PR, I tweaked the behavior of PUT requests to /user/events/favorited and /user/events/notified so that they ignore invalid items, and include an array in their response saying which items were actually added.

This also involved some major test rewrites because I had to change the parameterized tests, which led to me finding out they had been broken and totally changing how they work.

I would especially appreciate feedback on the actual code changes (`src/backend/api/utils.cjs` and `src/backend/api/routes/user.cjs`), but if you have comments on the tests or the docs changes don't hesitate to bring it up.

### DIFF
--- a/docs/api-v1.yaml
+++ b/docs/api-v1.yaml
@@ -306,7 +306,7 @@ paths:
       security:
         - auth_token: []
       requestBody:
-        description: Array of new favorited event IDs.
+        description: Array of new favorited event IDs. Any nonexistent IDs will be ignored.
         content:
           application/json:
             schema:
@@ -319,7 +319,7 @@ paths:
                     type: integer
       responses:
         '200':
-          description: Successful update.
+          description: Successful update. Any nonexistent IDs will be ignored.
           content:
             application/json:
               schema:
@@ -328,6 +328,11 @@ paths:
                   message:
                     type: string
                     example: Successfully updated favorited_events.
+                  new_value:
+                    type: array
+                    example: [77]
+                    items:
+                      type: integer
         '400':
           description: Bad request.
           content:
@@ -377,7 +382,7 @@ paths:
       tags:
         - user
       summary: Change notified events list.
-      description: Overwrite the array of notified events with a new array.
+      description: Overwrite the array of notified events with a new array. Any nonexistent IDs will be ignored.
       security:
         - auth_token: []
       requestBody:
@@ -393,8 +398,8 @@ paths:
                   items:
                     type: integer
       responses:
-        '202':
-          description: Successful update.
+        '200':
+          description: Successful update. Any nonexistent IDs will be ignored.
           content:
             application/json:
               schema:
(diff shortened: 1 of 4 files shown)


docs/api-v1.yaml
```diff
@@ -306,7 +306,7 @@ paths:
       security:
         - auth_token: []
       requestBody:
-        description: Array of new favorited event IDs.
+        description: Array of new favorited event IDs. Any nonexistent IDs will be ignored.
```
AnthonyS25223, line 309: this is crucial thank you so much!

docs/api-v1.yaml
```diff
@@ -393,8 +398,8 @@ paths:
                   items:
                     type: integer
       responses:
-        '202':
-          description: Successful update.
+        '200':
```
AnthonyS25223, line 401: Just wondering why is this 200 instead of 202 now?

src/backend/api/routes/user.cjs
```diff
@@ -187,7 +187,7 @@ async function routeGetFavorited(req, res, _next) {
 async function routePutFavorited(req, res, next) {
   // Set the favorited_events array of that user, and let that function do the response.
   // This assumes the body has `favorited_events`, and the middleware makes sure that is true
-  await util.userDataSetArray('favorited_events', req, res, next);
+  await util.setUserEventArray('favorited_events', req, res, next);
```
AnthonyS25223, line 190: Way clearer naming convention

src/backend/api/utils.cjs
```diff
@@ -416,69 +416,51 @@ async function asyncFilter(array, predicate) {
...
+    const existingIDs = await filterExistingEvents(newArray);
 
     // Update that user's favorited events in the database. Turn the array into a
     // string so it will be handled properly by the DB query.
-    await database.modifyAccountField(email, array_name, JSON.stringify(newArray));
+    await database.modifyAccountField(email, array_name, JSON.stringify(existingIDs));
     res.json({
-        'message': `Successfully updated ${array_name}`
+        'message': `Successfully updated ${array_name}.`,
+        'new_value': existingIDs  // send back what we did insert, because it may not match what they sent
```
AnthonyS25223, line 463: Great note of documentation

AnthonyS25223 (approved):
This code looks great! Nicely done and thank you so much for your work!

src/backend/api/routes/user.cjs
almondheil: Yeah it was pretending to be general when it wasn't lol. It can only do events because it checks for valid IDs.

docs/api-v1.yaml
almondheil: Oh yeah I messed this up before. It's just 200. 202 means "Accepted" which is semantically weird because it doesn't imply that the operation worked, just that it's valid. 200 makes more sense here.

PR #142, opened 2025-05-07
Writing

michael-paulin:
Has sprint planning for sprint 5.


.vscode/settings.json
almondheil, line 1: I should really delete this, it's not relevant at all. Will do that before I approve.

almondheil (approved):
This all looks good!

I don't have specific comments on lines, since we all collaborated to get these done.

Thank you for making the PR Michael!

PR #146, opened 2025-05-07
Upload artifact when building docs

almondheil:
This PR adds a step to our build process for new docs, where it `zip`-s up all the files and makes them available for download by us. This may be helpful for exploring other deployment options.


.github/workflows/deploy_mkdocs.yaml
```diff
@@ -1,12 +1,13 @@
 name: Build and Deploy MkDocs
 
 on:
-  # Run on pushes to specified branches.
+  # Run on pushes to main (i.e. when we merge a PR)
```
AnthonyS25223, line 4: Very good note to know

.github/workflows/deploy_mkdocs.yaml
```diff
@@ -1,12 +1,13 @@
...
   push:
-    branches: [$default-branch]
+    branches: [main]
+    # Only run if the docs files have changed
```
AnthonyS25223, line 7: Good distinction

AnthonyS25223 (approved):
Overall this change to actions makes sense! Thank you for the change and the distinction in the comments

PR #153, opened 2025-05-10
Enable changing of email

almondheil:
# To reviewer

This PR will allow you to change your email address, which can happen if you change your name through the college.

As per usual for my PRs there are a few types of change I made.

- `src/backend/api/...`: Write these new routes and make them exist.
- `src/backend/tests/...`: Allow these methods to be autotested as much as possible.
- `test procedures/...`: Rewrite manual tests to use scripts instead, for easier testing.
- `docs/api-v1.yaml`: Change API YAML description for docs site

# To almond (right after merging)

This PR changes the structore of the local OTP database, and you'll need to update that.

You need to construct a node script that will run this command on the sqlite db.

```
ALTER TABLE data ADD COLUMN oldEmail TEXT
```

This is a one-time thing.


src/backend/api/utils.cjs
```diff
@@ -105,13 +105,64 @@ function validateUsername(username) {
...
+    // where it will send the login.
     sendTo = email
-    if (email.toLowerCase() == 'getgrinnected.demo@grinnell.edu'){
+    if (email.toLowerCase() == 'getgrinnected.demo@grinnell.edu') {
         sendTo = 'getgrinnected.demo@gmail.com'
     }
+
     // Send the user a one-time code by email.
-    const code = sendCode(sendTo);
+    const code = sendEmailCode(sendTo);
```
AnthonyS25223, line 165: Just wondering what the difference in distinction was here other than maybe to be slightly clearer what code is being sent?

src/backend/db_connect.js
```diff
@@ -321,6 +321,23 @@ async function deleteAccount(email) {
...
+    if (!shouldExist) {
+        throw new Error('No previous account with old email');
+    }
+    const shouldNotExist = await getAccountByEmail(newEmail);
+    if (shouldNotExist) {
+        throw new Error('Account already exists with new email')
+    }
+
+    // Actually perform the switch
+    const sql = `UPDATE accounts SET email = ? WHERE email = ?`;
```
AnthonyS25223, line 336: Very simple an clear query!

src/backend/api/utils.cjs
almondheil: I've had a long-standing set of functions with bad names, and before this commit it was
- util.sendOTP
- user.sendCode

Those are not really different from each other, and I'm trying to make it slightly more different but idk if it worked.

It's a thing where I need to refactor that stuff but I'm hesitant to because it would become a really long function.

PR #155, opened 2025-05-11
Mark all request bodies as required

almondheil:
This PR is small, it just adds `required: true` to all the request body entries because all our bodies are in fact required.

No rush to review, this one being merged or not won't prevent my work.


AnthonyS25223 (approved):
Good not to make the changes for the docs! I think this is good to go through!

PR #161, opened 2025-05-12
Fix Verification Bug, Fully Implement Change Email, Refactor Verification Flow, Fix Merge Conflict (Missing Snackbar)

AnthonyS25223:
- I handled the bug where if the user left the app to check their email when they were on the verification page it now keeps them on the verification page.
- The change email button now follows the intended flow and makes the connection and call with the api
- I refactored the verification flow to use a string for the previous page as opposed to a flag variable
- I added the snack bar back to the home screen, as it was just completely gone for some reason. I assume it was lost on a merge at some point.

### DIFF
--- a/src/android/GetGrinnected/MyApplication/app/src/main/java/com/GetGrinnected/myapplication/DataStoreSettings.kt
+++ b/src/android/GetGrinnected/MyApplication/app/src/main/java/com/GetGrinnected/myapplication/DataStoreSettings.kt
@@ -29,6 +29,10 @@ object DataStoreSettings {
     private val LAST_SYNC_TIME = longPreferencesKey("last_sync_time")
     // Preference key for the font size
     private val FONT_SIZE = stringPreferencesKey("font_size")
+    // Preference key for when a user is trying too do a verification step
+    private val PENDING_EMAIL = stringPreferencesKey("pending_email")
+    // Preference key for which page we came from to get to verification ie the signup flag
+    private val PENDING_FLAG = booleanPreferencesKey("pending_flag")
 
     /**
      * Setter for logged in value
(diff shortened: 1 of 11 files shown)


src/android/GetGrinnected/MyApplication/app/src/main/java/com/GetGrinnected/myapplication/DataStoreSettings.kt
```diff
@@ -176,4 +180,44 @@ object DataStoreSettings {
         }
     }
 
+    /**
```
EthanEdwinHughes, line 183: I'm assuming all the pending tags are used to hold the process while users swap to their email. Well documented

src/android/GetGrinnected/MyApplication/app/src/main/java/screens/EmailVerificationScreen.kt
```diff
@@ -74,6 +76,19 @@ fun EmailVerificationScreen(email: String, flag: Boolean, navController: NavCont
     // Boolean associated with specifically a code error to shift field color
     var errCode by remember { mutableStateOf(false) }
 
+    // This is the persisted email value to be reloaded if the user leaves the app to check their email
```
EthanEdwinHughes, line 79: This is a cool piece of code to maintain the state of the email

src/android/GetGrinnected/MyApplication/app/src/main/java/screens/HomeScreen.kt
```diff
@@ -130,224 +135,250 @@ fun HomeScreen(tags: List<Check>) {
...
-                .padding(horizontal = 8.dp),
-        )
-        {
-            //creates a row to align the logo and buttons on the home page
-            Row(
-                verticalAlignment = Alignment.CenterVertically,
-                horizontalArrangement = Arrangement.SpaceBetween,
-                modifier = Modifier.fillMaxWidth()
+
+    // Host for our snack bar
```
EthanEdwinHughes, line 139: I am a little lost on what the snackbar is used for or what it is doing

EthanEdwinHughes (approved):
looks good

src/android/GetGrinnected/MyApplication/app/src/main/java/com/GetGrinnected/myapplication/DataStoreSettings.kt
AnthonyS25223: That's correct, these pending states are used to hold verification related information so that they can properly do verification when they comeback in

src/android/GetGrinnected/MyApplication/app/src/main/java/screens/HomeScreen.kt
AnthonyS25223: This is really just for the little pop up that comes up when the user favorites an event that says "you favorite blank event"

PR #163, opened 2025-05-12
Cleanup session routes to own file

almondheil:
This PR moves the session routes to their own file, because I forgot to do that before.

All the diff lines are just moving code around, and the autotests pass, so I think we're looking pretty good.


src/backend/api/api.cjs
```diff
@@ -108,22 +109,22 @@ function run() {
...
+    session.routeSendOTP
   );
   app.post(
     '/session/signup',
     [
       middlewareBodyExists('email'),
       middlewareBodyExists('username'),
     ],
-    user.routeSignUpNewUser
+    session.routeSignUpNewUser
```
AnthonyS25223, line 120: This makes a lot more sense for refactoring the location to be a session as opposed to a user

src/backend/api/routes/session.cjs
```diff
@@ -0,0 +1,191 @@
...
+    if (!await db.getAccountByEmail(email)) {
+        res.status(404).json({
+            'error': 'No such user',
+            'message': 'No user account exists with that email.'
+        });
+        return;
+    }
+
+    // Send that user an OTP code, then respond we did it successfully!
+    // TODO: we may want sendOTP to return error and deal with that.
```
AnthonyS25223, line 27: Is this still a TODO? just wondering

AnthonyS25223 (commented):
Overall this looks like a great way to refactor the location of the code to make more sense for where you would find these calls. Thank you so much for the amazing work it looks great!

AnthonyS25223 (approved):
Forgot to add approve and did just a comment instead

src/backend/api/routes/session.cjs
almondheil: It is, yes. This might just sorta be here for the forseeable future, it feels dishonest to remove the TODO without doing the thing.

Doing that thing hasn't been necessary yet.

PR #169, opened 2025-05-12
Change name order on settings pages

almondheil:
This PR puts our names in the alphabetical order by first name on both settings pages.

* almond Heil
* Anthony Schwindt
* Budhil Thijm
* Ellie Seehorn
* Ethan Hughes
* Michael Paulin

Before it was really unclear what the order was, and they were in a different order.

I'm open to a different ordering, but we want it to make sense and we want it to be the same across both apps. I don't especially like being at the front but that happens unless we sort `a` after `A-Z`, which Ellie reminds me is weird.

### DIFF
--- a/src/android/GetGrinnected/MyApplication/app/src/main/java/screens/SettingsScreen.kt
+++ b/src/android/GetGrinnected/MyApplication/app/src/main/java/screens/SettingsScreen.kt
@@ -553,12 +553,12 @@ fun SettingsScreen(modifier: Modifier = Modifier,
                             )
 
                             Text(
-                                text = "• Ellie Seehorn \n" +
-                                        "• Michael Paulin \n" +
+                                text = "• almond Heil \n" +
+                                        "• Anthony Schwindt \n" +
                                         "• Budhil Thijm \n" +
-                                        "• Almond Heil \n" +
+                                        "• Ellie Seehorn \n" +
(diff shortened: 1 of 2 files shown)


src/android/GetGrinnected/MyApplication/app/src/main/java/screens/SettingsScreen.kt
```diff
@@ -553,12 +553,12 @@ fun SettingsScreen(modifier: Modifier = Modifier,
                             )
 
                             Text(
-                                text = "• Ellie Seehorn \n" +
-                                        "• Michael Paulin \n" +
+                                text = "• almond Heil \n" +
```
seehorne, line 556: i know you don't like being first, but this makes the most sense i think

src/ios/GetGrinnected/GetGrinnected/CoreViews/SettingsView.swift
```diff
@@ -367,12 +367,12 @@ struct SettingsView: View {
                         
                     Text("Development Team")
```
seehorne, line 368: good to fix this in both apps

seehorne (approved):
Thank you for doing this, I know it is uncomfortable to be first, but it must be done on both first name and last, thank you for taking one for the team

PR #173, opened 2025-05-13
Maybe the real branch for that code

EthanEdwinHughes:
It cleans up and comments some code. Adjusts the UI of the Search Page to look nice as well as fixing date and location searching. Plus added a superficial notification time setting.

PR #174, opened 2025-05-13
Final Kotlin Code (Ethan)

EthanEdwinHughes:
I hate Git and everything else


almondheil (approved):
Okay we're good, thank you so much for your work.

I am sorry for the git pains.

To see original review (got put on wrong branch, but we're past it): #173

PR #176, opened 2025-05-13
Added a verification code alert when changing email.

michael-paulin:


src/ios/GetGrinnected/GetGrinnected/CoreViews/SettingsView.swift
```diff
@@ -196,11 +197,12 @@ struct SettingsView: View {
                 .transition(.asymmetric(
                     insertion: .move(edge: .top),
                     removal: .move(edge: .top)
-                ))//transition so that it pulls down instead of appearing out of nowhere
-//                    .transition(.move(edge: .top)) //alternative transition
+                ))//transition so that it pulls down instead of appearing out of nowhere=
                 
                 // username change
                 Button(action: {
+                    showVerificationAlert.toggle()
```
seehorne, line 204: ahh this exciting, thanks for adding

seehorne (approved):
looks great, thank you for adding!

src/ios/GetGrinnected/GetGrinnected/CoreViews/SettingsView.swift
```diff
@@ -267,13 +269,17 @@ struct SettingsView: View {
                     Image(systemName: "square.and.arrow.up")
                         .imageScale(.large)
                 } //Button
+                .alert("Enter verification code", isPresented: $showVerificationAlert) {
```
seehorne, line 272: ah this makes sense, i think when i had tried i had gotten placement weird and it hated it so i just moved on, but the logic is clear

src/ios/GetGrinnected/GetGrinnected.xcodeproj/xcuserdata/michael.xcuserdatad/xcdebugger/Breakpoints_v2.xcbkptlist
```diff
@@ -16,7 +16,7 @@
             endingColumnNumber = "9223372036854775807"
             startingLineNumber = "234"
             endingLineNumber = "234"
-            landmarkName = "updatePossibleTags(events:)"
+            landmarkName = "updateEvents()"
```
seehorne, line 19: commenting on this to make the # of comments right, this is xcuserdata so in general we don't need to push it but it also doesn't hurt much, so in the future if there is a future in your commits (you might be done and if so congrats), you can refrain but its like not a problem

PR #177, opened 2025-05-13
Implement check for changing to same email

almondheil:
I really don't know how this got past both me and the original reviewer, but thanks to Ellie we found this bug and I have a pretty robust fix for it now.

I have also added tests here to make sure if this happens again, we'll know.

### DIFF
--- a/src/backend/api/routes/user.cjs
+++ b/src/backend/api/routes/user.cjs
@@ -181,6 +181,15 @@ async function routeChangeEmail(req, res, _next) {
   const newEmail = req.body.new_email;
   const oldEmail = req.email;
 
+  // Make sure the emails are not the same, space- and case-insensitively.
+  if (newEmail.trim().toLowerCase() === oldEmail.trim().toLowerCase()) {
+    res.status(400).json({
+      'error': 'Same email',
+      'message': 'The old and new emails must be different addresses.'
+    });
+    return;
+  }
+
   // Reject the response if the email isn't good.
   const valid = util.validateEmail(newEmail);
   if (!valid.result) {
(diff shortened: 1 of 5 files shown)


src/backend/api/routes/user.cjs
```diff
@@ -181,6 +181,15 @@ async function routeChangeEmail(req, res, _next) {
   const newEmail = req.body.new_email;
   const oldEmail = req.email;
 
+  // Make sure the emails are not the same, space- and case-insensitively.
+  if (newEmail.trim().toLowerCase() === oldEmail.trim().toLowerCase()) {
```
seehorne, line 185: good check!

src/backend/test/test_api.js
```diff
@@ -568,11 +565,74 @@ describe('Test API', () => {
                 assert.strictEqual(res.statusCode, 200, res.text);
                 assert.strictEqual(
                     res.text,
-                    '{"email":"email@example.com"}'
+                    '{"email":"example1234567890@grinnell.edu"}'
                 );
             });
         });
 
+        /*
```
seehorne, line 573: thank you for this test, this will be super valuable

src/backend/test/test_api.js
```diff
@@ -568,11 +565,74 @@ describe('Test API', () => {
...
+                    .send({ 'new_email': 'example1234567890@grinnell.edu' });
+
+                assert.strictEqual(res.statusCode, 400, res.text);
+                assert.strictEqual(res.text, JSON.stringify({
+                    'error': 'Same email',
+                    'message': 'The old and new emails must be different addresses.'
+                }));
+            });
+
+            it("rejects changing to the same email but capitalized", async () => {
```
seehorne, line 621: perfect! some people if they have autocaps on will have an issue with this happening unintentionally so good to guard against it and have a test demonstrating we do so

seehorne (approved):
This is very helpful, thank you for updating this

PR #179, opened 2025-05-13
Create script to delete expired one-time codes

almondheil:
This PR creates a script that deletes any expired one-time codes from the sqlite3 database, which stops them from piling up if people do not verify them.

I have tested it with both expired and non-expired codes on my machine, but no automatic testing of this component is planned right now.

When it is merged, I can add a cron job on the server that runs this script every 12 hours or something.

### DIFF
--- a/src/backend/clean_otp_codes.cjs
+++ b/src/backend/clean_otp_codes.cjs
@@ -1,6 +1,14 @@
 const sqlite3 = require('sqlite3');
 const DBPATH = './src/backend/Database/localOTP.db'
 
+/**
+ * Clean any expired OTP codes in the specified filename. Codes will be cleaned up
+ * if their expiration time is less than the current time when this function is called.
+ * 
+ * If any error happens with the database, will print that error to the screen.
+ * 
+ * @param {string} filename Filename of sqlite database to target.
+ */
 async function otpFileClean(filename) {
     // Format the current time in the same format as the database
     //   example: 2025-05-13T18:51:51.899Z


src/backend/clean_otp_codes.cjs
```diff
@@ -0,0 +1,29 @@
+const sqlite3 = require('sqlite3');
+const DBPATH = './src/backend/Database/localOTP.db'
+
+async function otpFileClean(filename) {
```
michael-paulin, line 4: Maybe add a header comment that says what this function does?

src/backend/clean_otp_codes.cjs
```diff
@@ -0,0 +1,29 @@
...
+    // Format the current time in the same format as the database
+    //   example: 2025-05-13T18:51:51.899Z
```
michael-paulin, line 6: Very helpful comments!

src/backend/clean_otp_codes.cjs
```diff
@@ -0,0 +1,29 @@
...
+    db.get(sql, now,
+        // callback function for the db command. print the error if we hit one.
+        (err, _) => {
+            if (err) {
+                console.error(err);
+            }
+        });
+}
+
+// when this file is run act like a script, but export our function otherwise
```
michael-paulin, line 24: What does this mean?

michael-paulin (changes requested):
Looks good! Just would like a header comment on the function that says what it does.

src/backend/clean_otp_codes.cjs
almondheil: This is the difference between "behavior when imported" and "behavior when run on its own".

This file is mainly intended to be run on its own.

Would you like me to change the comment to clarify that?

src/backend/clean_otp_codes.cjs
almondheil: Yeah, for sure!

michael-paulin: Ok new header comment looks great!

michael-paulin (approved):
Looks great! You can merge now.

PR #183, opened 2025-05-13
Hotfix: tags button color contrast was broken

seehorne:
one line change, don't delete the branch after. but it makes the tags label have better contrast


src/ios/GetGrinnected/GetGrinnected/Components/TagMultiSelector.swift
```diff
@@ -45,7 +45,7 @@ struct TagMultiSelector: View {
             Label(title, systemImage: "chevron.down")
                 .padding(.all, 3)
                 .font(.subheadline)
-                .foregroundColor(.white)
+                .foregroundColor(Color.appBorder)
```
almondheil, line 48: Yep that's a pretty good one-line change, to remove the hardcoding

almondheil (approved):
Like you said almost no changes, I think all good.

